PR #17, opened 2019-05-29
Try to auto-fix service name constraints

### LINKED ISSUE #9: Service Name Contraints
See: https://github.com/jamesward/cloud-run-button/issues/4

ahmetb:
Fixes #9.

cc: @steren @jamesward

### DIFF
--- a/cmd/cloudshell_open/cloudrun.go
+++ b/cmd/cloudshell_open/cloudrun.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"os/exec"
 	"strings"
+	"unicode"
 )
 
 func deploy(project, name, image, region string, envs []string) (string, error) {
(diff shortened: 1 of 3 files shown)


cmd/cloudshell_open/cloudrun.go
```diff
@@ -46,3 +47,26 @@ func serviceURL(project, name, region string) (string, error) {
...
+// tryFixServiceName attempts replace the service name with a better one to
+// prevent deployment failures due to Cloud Run service naming constraints such
+// as:
+//
+//   * names with a leading non-letter (e.g. digit or '-') are prefixed
+//   * names over 63 characters are truncated
+//   * names ending with a '-' have the suffix trimmed
+func tryFixServiceName(name string) string {
+	if name == "" {
+		return name
```
jamesward: Might be missing some context elsewhere but an empty string shouldn't be a valid name.

ahmetb: sure, this just prevents the function to `panic` on bad input. it's not valid input, but shouldn't cause the program to crash.

jamesward: Should we provide a default in this case?  Or maybe we do elsewhere.

ahmetb: I think this is fine since "" is not supposed to happen. But when it happens, it doesn't crash the program.

There are certain safeguards below that if we end up with empty service name we return `svc`. If you follow the code below, e.g. `-----` is returned as `svc`.

PR #30, opened 2019-05-31
Wait for active billing account

### LINKED ISSUE #24: Handling When Billing Not Enabled
Like:
```
  gcloud services enable run.googleapis.com 2> /dev/null

  if [[ $? -eq 1 ]]; then
    echo "Your account needs to have billing setup to continue. Visit:"
    echo "https://console.developers.google.com/billing/linkedaccount?project=$PROJECT_ID\n"
    read -p "After setting up billing, press enter to continue."
    gcloud services enable run.googleapis.com
  fi
```

ahmetb:
Using "gcloud beta billing projects describe --format=json" to determine if
billing is enabled. Hopefully this won't break anytime soon, but we should
move to GA version of the cmd once done.

Fixes #24

### DIFF
--- a/cmd/cloudshell_open/billing.go
+++ b/cmd/cloudshell_open/billing.go
@@ -1,0 +1,42 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


ahmetb:
Looks like this:

image

PR #44, opened 2019-06-18
use buildpacks when Dockerfile is not present

### LINKED ISSUE #43: Add buildpack support
If there is no Dockerfile, fallback to trying buildpacks.  See #3 for an add-on / related discussion.

poy:
This CL uses the heroku/buildpacks builder if a Dockerfile is not found.

fixes #43

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,13 @@ RUN go mod download
 COPY . .
 RUN CGO_ENABLED=0 GOOS=linux go build -a -ldflags '-extldflags "-static"' \
         -o /bin/a.out ./cmd/cloudshell_open
+WORKDIR /tmp
+RUN wget https://github.com/buildpack/pack/releases/download/v0.2.1/pack-v0.2.1-linux.tgz && tar -xzf pack-v0.2.1-linux.tgz
+RUN wget https://raw.githubusercontent.com/buildpack/pack/v0.2.1/LICENSE
 
 FROM gcr.io/cloudshell-images/cloudshell:latest
 RUN rm /google/devshell/bashrc.google.d/cloudshell_open.sh
 COPY --from=build /bin/a.out /bin/cloudshell_open
+COPY --from=build /tmp/pack /opt/pack/pack
+COPY --from=build /tmp/LICENSE /opt/pack/LICENSE
+RUN ln -s /opt/pack/pack /usr/local/bin/pack
--- a/cmd/cloudshell_open/docker.go
+++ b/cmd/cloudshell_open/docker.go
@@ -16,10 +16,12 @@ package main
 
 import (
 	"fmt"
+	"os"
 	"os/exec"
+	"path/filepath"
 )
 
-func build(dir, image string) error {
+func dockerBuild(dir, image string) error {
 	cmd := exec.Command("docker", "build", "--quiet", "--tag", image, dir)
 	b, err := cmd.CombinedOutput()
 	if err != nil {
@@ -28,11 +30,22 @@ func build(dir, image string) error {
 	return nil
 }
 
-func push(image string) error {
+func dockerPush(image string) error {
 	cmd := exec.Command("docker", "push", image)
 	b, err := cmd.CombinedOutput()
 	if err != nil {
 		return fmt.Errorf("docker push failed: %v, output:\n%s", err, string(b))
 	}
 	return nil
+}
+
+func dockerFileExists(dir string) (bool, error) {
+	if _, err := os.Stat(filepath.Join(dir, "Dockerfile")); err != nil {
+		if os.IsNotExist(err) {
+			return false, nil
+		}
+		return false, fmt.Errorf("failed to check for Dockerfile in the repo: %v", err)
+	}
+
+	return true, nil
 }
--- a/cmd/cloudshell_open/main.go
+++ b/cmd/cloudshell_open/main.go
@@ -214,16 +214,30 @@ func run(c *cli.Context) error {
 	serviceName = tryFixServiceName(serviceName)
 
 	image := fmt.Sprintf("gcr.io/%s/%s", project, serviceName)
-	fmt.Println(infoPrefix + " FYI, running the following command:")
-	cmdColor.Printf("\tdocker build -t %s %s\n", parameter(image), parameter("."))
 
 	end = logProgress(fmt.Sprintf("Building container image %s", highlight(image)),
 		fmt.Sprintf("Built container image %s", highlight(image)),
 		"Failed to build container image.")
-	err = build(appDir, image)
-	end(err == nil)
+
+	exists, err := dockerFileExists(appDir)
 	if err != nil {
 		return err
+	}
+
+	if exists {
+		fmt.Println(infoPrefix + " Attempting to build this application with its Dockerfile...")
+		fmt.Println(infoPrefix + " FYI, running the following command:")
+		cmdColor.Printf("\tdocker build -t %s %s\n", parameter(image), parameter("."))
+		err = dockerBuild(appDir, image)
+	} else {
+		fmt.Println(infoPrefix + " Attempting to build this application with Cloud Native Buildpacks (buildpacks.io)...")
+		fmt.Println(infoPrefix + " FYI, running the following command:")
+		cmdColor.Printf("\tpack build %s --path %s --builder heroku/buildacks\n", parameter(image), parameter(appDir))
+		err = packBuild(appDir, image)
+	}
+	end(err == nil)
+	if err != nil {
+		return fmt.Errorf("this application doesn't have a Dockerfile; attempted to build it via heroku/buildpacks and failed: %s", err)
 	}
 
 	fmt.Println(infoPrefix + " FYI, running the following command:")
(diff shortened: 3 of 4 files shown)


jamesward:
This is great! Think we should have any tests around this?

cmd/cloudshell_open/build.go
```diff
@@ -0,0 +1,38 @@
...
+
+		if err := dockerPush(image); err != nil {
+			return fmt.Errorf("failed to push image to %s: %+v", image, err)
+		}
+
+		return nil
+	}
+
+	// Use buildpacks
+	return packBuildAndPublish(dir, image)
```
ahmetb: See my comment in the relevant issue. 
I think we should not fall back, we should try this path ONLY if a particular file exists. 

Similarly the error shouldnt say “pack failed” IF there’s no dockerfile in dir. it should say “no dockerfile or procfile”. So i think you need a procfileExists() here.

cmd/cloudshell_open/docker.go
```diff
@@ -28,11 +33,28 @@ func build(dir, image string) error {
 	return nil
 }
 
-func push(image string) error {
+func dockerPush(image string) (err error) {
+	fmt.Println(infoPrefix + " FYI, running the following command:")
+	cmdColor.Printf("\tdocker push %s\n", parameter(image))
+	end := logProgress("Pushing container image...",
```
ahmetb: Lets not move printing code in here. It gets really spaghetti. 

Instead we can say “Building & pushing” in main().

cmd/cloudshell_open/docker.go
```diff
@@ -28,11 +33,28 @@ func build(dir, image string) error {
...
 	cmd := exec.Command("docker", "push", image)
 	b, err := cmd.CombinedOutput()
 	if err != nil {
 		return fmt.Errorf("docker push failed: %v, output:\n%s", err, string(b))
 	}
 	return nil
 }
+
+func dockerFileExists(dir string) bool {
+	if _, err := os.Stat(filepath.Join(dir, "Dockerfile")); err != nil {
```
ahmetb: Err != nil is a bad check here
This  can fail for other reasons. Need to test:
1. Is file
2. Err is ENOENT.

cmd/cloudshell_open/docker.go
ahmetb: Feels like we should move the logic in build() to main()

This way we can keep logging stuff there too

poy: I think the main should be a little thinner TBH. Its near impossible to test as is.

ahmetb:
Thanks for this!
Mostly lgtm, mainly my concern in #43 and some cosmetics.

cmd/cloudshell_open/build.go
jamesward: Procfile isn't required for buildpacks.  They are really designed to have nothing special in the repo to make them buildpack compatible.  So I think it makes sense to fallback to buildpacks since buildpacks have a detect phase which will fail if it can't build the repo.

cmd/cloudshell_open/docker.go
poy: I think logging in most Go projects is buck shot throughout the project by passing a "logger" of some sort (e.g., `log.Logger`).

In this case, as everything is in the main package and utilizing globals, I think the real problem is that the different functions need to be extracted. I figured that was outside the scope of this PR.

cmd/cloudshell_open/docker.go
poy: Done

cmd/cloudshell_open/build.go
ahmetb: But the primary path is dockerfile. We don’t wanna fail with buildpack error and totally don’t mention dockerfile when fallback fails, too. 

My experimentation with heroku buildpacks showed me Procfile is required. Is this totally incorrect?

cmd/cloudshell_open/docker.go
ahmetb: don’t worry about testing it, as long as we keep the logic there trivial. 

It would be nice if the path we choose for building is trivial and doesn’t require testing. My main concern is moving the UI code out of main() where it’s less discoverable.

cmd/cloudshell_open/build.go
jamesward: Since I don't know Go I can't totally tell but I think the experience should be something like:
```
if Dockerfile.exists then doDockerBuild # if this fails stop and display the error
else echo "Dockerfile not found, trying buildpacks"; doBuildpackBuild # if this fails stop and display the error
```

The `Procfile` can often be inferred by the buildpack.  But in cases when it can't, or when something custom is needed, it can be provided as an override.

cmd/cloudshell_open/docker.go
poy: I inlined most of this along with the logging.

cmd/cloudshell_open/build.go
poy: @ahmetb I got this working with Go for https://github.com/poy/knative-testing

Dockerfile
```diff
@@ -6,7 +6,11 @@ RUN go mod download
 COPY . .
 RUN CGO_ENABLED=0 GOOS=linux go build -a -ldflags '-extldflags "-static"' \
         -o /bin/a.out ./cmd/cloudshell_open
+WORKDIR /tmp
+RUN wget https://github.com/buildpack/pack/releases/download/v0.2.1/pack-v0.2.1-linux.tgz && tar -xzf pack-v0.2.1-linux.tgz
 
 FROM gcr.io/cloudshell-images/cloudshell:latest
 RUN rm /google/devshell/bashrc.google.d/cloudshell_open.sh
+RUN apt update && apt install docker-ce -y
```
ahmetb: ideally we shouldn't install `docker` in this image, as it should be present in the cloudshell base image already.
if it's not the right version or something like that, we should probably consider updating it. having two versions installed would probably cause an issue at some point later on.

ahmetb:
this is lgtm
hold until we figure out how we can ship `pack` CLI in the shell image.

Dockerfile
poy: This should be resolved soon

jamesward:
Going through some legal review stuff, can we move `pack` to `/opt/pack` and symlink it into `/usr/local/bin` and also include the [license](https://github.com/buildpack/pack/blob/master/LICENSE) in that dir?

Dockerfile
jamesward: We do just need a new version and that can be done instead with `apt upgrade docker-ce`

ahmetb: Let's get this fixed in cloudshell base image.
Here we're updating to latest version that we don't know whether is good or bad. At least we can follow cloudshell's release cycles.

poy: Looks like it has been. We'll just have to wait a little bit for it to get rolled out. I've removed the line.

poy:
@jamesward 
> Going through some legal review stuff, can we move pack to /opt/pack/pack and symlink it into /usr/local/bin and also include the license in the /opt/pack dir?

done

poy:
> This is great! Think we should have any tests around this?

Discussed with @jamesward offline, we decided to punt on tests for the moment until the project is pretty structured for them.

jamesward:
> @jamesward
> 
> > Going through some legal review stuff, can we move pack to /opt/pack/pack and symlink it into /usr/local/bin and also include the license in the /opt/pack dir?
> 
> done

I updated my comment after you saw it. :)  Can you sub-dir it and put the pack license in that subdir?

poy:
> I updated my comment after you saw it. :) Can you sub-dir it and put the pack license in that subdir?

@jamesward Seems more likely that I didn't read it very carefully :)

Updated. PTAL

cmd/cloudshell_open/main.go
```diff
@@ -214,24 +214,36 @@ func run(c *cli.Context) error {
...
+		cmdColor.Printf("\tdocker build -t %s %s\n", parameter(image), parameter("."))
+		err = dockerBuild(appDir, image)
+	} else {
+		fmt.Println(infoPrefix + " FYI, running the following command:")
+		cmdColor.Printf("\tpack build %s --path %s --builder heroku/buildacks\n", parameter(image), parameter(appDir))
+		err = packBuild(appDir, image)
+	}
+	end(err == nil)
+	if err != nil {
+		return fmt.Errorf("failed to build image: %s", err)
```
ahmetb: can we do this:

    this application doesn't have a Dockerfile, attempted to build it as a buildpack and failed: %v

poy: PTAL

cmd/cloudshell_open/main.go
```diff
@@ -214,24 +214,36 @@ func run(c *cli.Context) error {
...
 	if err != nil {
 		return err
 	}
 
+	if exists {
+		fmt.Println(infoPrefix + " FYI, running the following command:")
+		cmdColor.Printf("\tdocker build -t %s %s\n", parameter(image), parameter("."))
+		err = dockerBuild(appDir, image)
+	} else {
+		fmt.Println(infoPrefix + " FYI, running the following command:")
```
ahmetb: wdyt of adding an always-visible info here when this path is taken:

    fmt.Println(infoPrefix + " Attempting to build this application with Cloud Native Buildpacks (buildpacks.io)...")

we can also add the same info to the dockerBuild block too

    fmt.Println(infoPrefix + " Attempting to build this application with its Dockerfile...")

jamesward:
This hasn't rolled out yet, but when it does, this should work:
https://console.cloud.google.com/cloudshell/editor?shellonly=true&cloudshell_image=gcr.io/cloudrun/button&cloudshell_git_repo=https://github.com/jamesward/hello-go.git&cloudshell_git_branch=buildpack

jamesward:
This is now working!

PR #55, opened 2019-07-17
Add prompt for region

ikasamt:
Cloud Run currently supports multiple regions. I added a prompt to select a region.

### DIFF
--- a/cmd/cloudshell_open/locations.go
+++ b/cmd/cloudshell_open/locations.go
@@ -1,0 +1,32 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/AlecAivazis/survey/v2"
+)
+
+func listLocations() ([]string, error) {
+	loc := []string{
+		`us-central1`,
+		`asia-northeast1`,
+		`euro-west1`,
+		`us-east1`,
+	}
+	return loc, nil
+}
+
+func promptLocation(locations []string) (string, error) {
+	var p string
+	if err := survey.AskOne(&survey.Select{
+		Message: "Choose a region to deploy this application:",
+		Options: locations,
+	}, &p,
+		surveyIconOpts,
+		survey.WithValidator(survey.Required),
+	); err != nil {
+		return p, fmt.Errorf("could not choose a region: %+v", err)
+	}
+	return p, nil
+
+}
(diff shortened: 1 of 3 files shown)


ahmetb:
By design we currently want this flow to be simpler and requiring minimal user input.  

That’s why we didn’t add region selection. 

Do you need this feature, as it will come at a cost to allusers. 

Also please next time send a proposal first instead of writing the code which might get rejected altogether.

cmd/cloudshell_open/locations.go
```diff
@@ -0,0 +1,32 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/AlecAivazis/survey/v2"
+)
+
+func listLocations() ([]string, error) {
+	loc := []string{
```
ahmetb: We shouldnt hardcode. There’s a locations api on cloud run. But that will require bringing in GCP go libraries.

jamesward: Can we get them from `gcloud` or via REST?

jamesward:
There are a number of other defaults that might be nice to expose (mem settings, concurrency, etc) so to keep the usability simple but allow overrides we could do something like:
```
Using defaults:
Region = us-central-1
Mem = 512MB
Concurrency = 80
Press space to change these.  Continuing with the defaults in 5...4...3...2..1
```

cmd/cloudshell_open/locations.go
ahmetb: I don't think so + it would be slower.
Similarly we can also get the projects via API once we import the package.

ahmetb:
> ```
> Press space to change these.  Continuing with the defaults in 5...4...3...2...1
> ```

This is kind of hard to do, and not a common pattern to do deadline-orinted prompts.

I prefer we defer mem/concurrency/CPU to the developer (owner repo, via app.json instead) as they know what their app requires.

Region is the deploying user's preference, so maybe we can make it a required prompt with a default value selected already (that users can just hit Enter) 🤷‍♂

jamesward:
Yeah, that is true.  I like that better.

ahmetb:
I have a PR at #62 that does this with the REST API (via the official client library). I’ll close this for now (but thanks for your PR!).

PR #60, opened 2019-07-22
document the cloudshell_git_branch param

jamesward:


README.md
```diff
@@ -25,6 +25,8 @@ Cloud](https://storage.googleapis.com/cloudrun/button.svg)](https://console.clou
 1. Make sure the repository has a Dockerfile, so it can be built using the
    `docker build` command.
 
+1. *Optional* To use a different git branch, add a `cloudshell_git_branch=BRANCH_NAME` query parameter
```
ahmetb: I think this deserves a new section, like `Customizing source repository` where you can talk about both cloudshell_git_branch, as well as cloudshell_dir (also undocumented).

Since most people won't need to customize these two, I'm inclined to have them outside the Instructions section relevant to ~every user.

jamesward: New section added

ahmetb:
This likely will be obsolete with redirector (as it can infer the HEAD name and subdirectory from github repos).

jamesward:
I think we will need to maintain the ability to override the referer stuff (like when a button is not on a github readme).

jamesward:
But we can definitely change param names with the redirector.

PR #69, opened 2019-08-16
use run v1alpha1 api

### LINKED ISSUE #68: Repo calling v1beta1 that doesn't exist
v1beta1 is called (for example here: https://github.com/GoogleCloudPlatform/cloud-run-button/blob/master/cmd/cloudshell_open/cloudrun.go) However, it doesn't exist as a 404 error is returned. Also, it's not mentioned in the public documentation (https://cloud.google.com/run/docs/reference/rest/) nor the Google API repos (https://github.com/googleapis/google-api-go-client/tree/master/cloudbuild)

ahmetb:
Previously I opted in to use v1beta1 (because well, Cloud Run is beta) but it
turns out v1 and v1beta1 apis are accidentally published in API Discovery
system and doesn't have public access. Reverting to v1alpha1.

Fixes #68.

### DIFF
--- a/cmd/cloudshell_open/cloudrun.go
+++ b/cmd/cloudshell_open/cloudrun.go
@@ -23,7 +23,7 @@ import (
 	"unicode"
 
 	"github.com/AlecAivazis/survey/v2"
-	runapi "google.golang.org/api/run/v1beta1"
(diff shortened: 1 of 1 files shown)


ahmetb:
I still haven't manually tested this, but will update shortly.

PR #72, opened 2019-08-22
Correct spelling on replication command

glasnt:
I don't believe this is a bug, just if a user tries to copy the sample command they will encounter issues due to the misspelling


jamesward:
Nice find!  Thanks!

PR #73, opened 2019-08-22
update demo image

ahmetb:
Higher resolution + had to change "courier new" to "Roboto Mono" in the
Chrome developer tools for stronger font body rendering.

@jamesward


jamesward:
Looks really good! We do also need to overlay a play button on it.  Or create another one in the readme with a play button so that people know they can click to watch a gif.

PR #82, opened 2019-09-04
add error handling

toshi0607:
* add error handling
* gofmt 3 files

### DIFF
--- a/cmd/cloudshell_open/main.go
+++ b/cmd/cloudshell_open/main.go
@@ -317,12 +317,16 @@ func checkCloudShellTrusted() (bool, error) {
 }
 
 func waitForBilling(projectID string, prompt func(string) error) error {
-	ok, err := checkBillingEnabled(projectID)
-	if err != nil {
-		return err
+	for {
+		ok, err := checkBillingEnabled(projectID)
+		if err != nil {
+			return err
+		}
+		if ok {
+			return nil
+		}
+		if err := prompt(projectID); err != nil {
+			return err
+		}
 	}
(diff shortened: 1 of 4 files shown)


cmd/cloudshell_open/main.go
```diff
@@ -311,15 +316,13 @@ func checkCloudShellTrusted() (bool, error) {
 	return false, fmt.Errorf("error determining if cloud shell is trusted: %+v. output=\n%s", err, string(b))
 }
 
-func waitForBilling(projectID string, prompt func(string)) error {
-	for {
```
ahmetb: looks like you've removed the for-loop here. why?

toshi0607: I thought this was unintentional, but is this intentional for waiting until billing will be enabled and is infinite loop OK if billing will not be enabled?

toshi0607: If this loop is intentional, I will revert loop removal and return result of prompt only when error happened, of course!

ahmetb: Yeah this is intentional. It would be good to preserve the logic.

toshi0607: Thank you! I'll fix!

ahmetb:
Thanks, looks good.

I'm not entirely sure how the error returned error will be used in this case. I suspect we'll start seeing error messages like `EOF` or `interrupt` due to these when user hits Ctrl+C or something like that.

At least we're not ignoring it now.

PR #86, opened 2019-09-24
add support for postdeploy script - fixes #76

### LINKED ISSUE #76: Post Deploy Scripts
Add support for: https://devcenter.heroku.com/articles/app-json-schema#scripts

jamesward:
I tested this with the `master`, `echo`, and `external` branches on:
https://github.com/jamesward/cloud-run-button-postdeploy-test

I don't know Go, so this PR might need a lot of help.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,10 +92,10 @@ Reference:
 - `options`: _(optional)_ Options when deploying the service
   - `allow-unauthenticated`: _(optional, default: `true`)_ allow unauthenticated requests
 - `hooks`: _(optional)_ Run scripts at phases of the deploy process
-  - `precreate`: _(optional) Runs the specified commands before the service has been created
-    - `commands`: _(array of strings) The list of commands to run
-  - `postcreate`: _(optional) Runs the specified commands after the service has been created
-    - `commands`: _(array of strings) The list of commands to run
+  - `precreate`: _(optional)_ Runs the specified commands before the service has been created
+    - `commands`: _(array of strings)_ The list of commands to run
+  - `postcreate`: _(optional)_ Runs the specified commands after the service has been created
+    - `commands`: _(array of strings)_ The list of commands to run
 
 ### Notes
 
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ Reference:
   - `generator`, _(optional)_ use a generator for the value, currently only support `secret`
 - `options`: _(optional)_ Options when deploying the service
   - `allow-unauthenticated`: _(optional, default: `true`)_ allow unauthenticated requests
-- `hooks`: _(optional)_ Run scripts at phases of the deploy process
+- `hooks`: _(optional)_ Run commands in separate bash shells with the environment variables configured for the application and environment variables for `PROJECT_ID`, `SERVICE`, and `REGION`. Command outputs are shown as they are executed.
   - `precreate`: _(optional)_ Runs the specified commands before the service has been created
     - `commands`: _(array of strings)_ The list of commands to run
   - `postcreate`: _(optional)_ Runs the specified commands after the service has been created
--- a/README.md
+++ b/README.md
@@ -91,7 +91,9 @@ Reference:
   - `generator`, _(optional)_ use a generator for the value, currently only support `secret`
 - `options`: _(optional)_ Options when deploying the service
   - `allow-unauthenticated`: _(optional, default: `true`)_ allow unauthenticated requests
-- `hooks`: _(optional)_ Run commands in separate bash shells with the environment variables configured for the application and environment variables for `PROJECT_ID`, `K_SERVICE`, and `REGION`. Command outputs are shown as they are executed.
+- `hooks`: _(optional)_ Run commands in separate bash shells with the environment variables configured for the
+ application and environment variables for `GOOGLE_CLOUD_PROJECT`, `GOOGLE_CLOUD_REGION`, and `K_SERVICE`. Command
+ outputs are shown as they are executed.
   - `precreate`: _(optional)_ Runs the specified commands before the service has been created
     - `commands`: _(array of strings)_ The list of commands to run
   - `postcreate`: _(optional)_ Runs the specified commands after the service has been created
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ Reference:
   - `generator`, _(optional)_ use a generator for the value, currently only support `secret`
 - `options`: _(optional)_ Options when deploying the service
   - `allow-unauthenticated`: _(optional, default: `true`)_ allow unauthenticated requests
-- `hooks`: _(optional)_ Run commands in separate bash shells with the environment variables configured for the application and environment variables for `PROJECT_ID`, `SERVICE`, and `REGION`. Command outputs are shown as they are executed.
+- `hooks`: _(optional)_ Run commands in separate bash shells with the environment variables configured for the application and environment variables for `PROJECT_ID`, `K_SERVICE`, and `REGION`. Command outputs are shown as they are executed.
   - `precreate`: _(optional)_ Runs the specified commands before the service has been created
     - `commands`: _(array of strings)_ The list of commands to run
   - `postcreate`: _(optional)_ Runs the specified commands after the service has been created
--- a/README.md
+++ b/README.md
@@ -62,8 +62,17 @@ For example:
     "options": {
         "allow-unauthenticated": false
     },
-    "hook": {
-        "postcreate": "postcreate.sh"
+    "hooks": {
+        "precreate": {
+            "commands": [
+                "echo 'test'"
+            ]
+        },
+        "postcreate": {
+            "commands": [
+                "./setup.sh"
+            ]
+        }
     }
 }
 ```
@@ -82,8 +91,11 @@ Reference:
   - `generator`, _(optional)_ use a generator for the value, currently only support `secret`
 - `options`: _(optional)_ Options when deploying the service
   - `allow-unauthenticated`: _(optional, default: `true`)_ allow unauthenticated requests
-- `hook`: _(optional)_ Run scripts at phases of the deploy process
-  - `postcreate`: _(optional) Runs the specified script file after the service has been created
+- `hooks`: _(optional)_ Run scripts at phases of the deploy process
+  - `precreate`: _(optional) Runs the specified commands before the service has been created
+    - `commands`: _(array of strings) The list of commands to run
+  - `postcreate`: _(optional) Runs the specified commands after the service has been created
+    - `commands`: _(array of strings) The list of commands to run
 
 ### Notes
 
--- a/README.md
+++ b/README.md
@@ -61,6 +61,9 @@ For example:
     },
     "options": {
         "allow-unauthenticated": false
+    },
+    "hook": {
+        "postcreate": "postcreate.sh"
     }
 }
 ```
@@ -79,6 +82,8 @@ Reference:
   - `generator`, _(optional)_ use a generator for the value, currently only support `secret`
 - `options`: _(optional)_ Options when deploying the service
   - `allow-unauthenticated`: _(optional, default: `true`)_ allow unauthenticated requests
+- `hook`: _(optional)_ Run scripts at phases of the deploy process
+  - `postcreate`: _(optional) Runs the specified script file after the service has been created
 
 ### Notes
 
--- a/cmd/cloudshell_open/appfile.go
+++ b/cmd/cloudshell_open/appfile.go
@@ -40,14 +40,19 @@ type options struct {
 }
 
 type hook struct {
-	PostCreate string `json:"postcreate"`
+	Commands []string `json:"commands"`
+}
+
+type hooks struct {
+	PreCreate hook `json:"precreate"`
+	PostCreate hook `json:"postcreate"`
 }
 
 type appFile struct {
 	Name        string         `json:"name"`
 	Env         map[string]env `json:"env"`
 	Options     options        `json:"options"`
-	Hooks       hook           `json:"hook"`
+	Hooks       hooks          `json:"hooks"`
 
 	// The following are unused variables that are still silently accepted
 	// for compatibility with Heroku app.json files.
--- a/cmd/cloudshell_open/appfile.go
+++ b/cmd/cloudshell_open/appfile.go
@@ -39,14 +39,18 @@ type options struct {
 	AllowUnauthenticated *bool `json:"allow-unauthenticated"`
 }
 
+type scripts struct {
+	Postdeploy string `json:"postdeploy"`
+}
+
 type appFile struct {
 	Name        string         `json:"name"`
 	Env         map[string]env `json:"env"`
+	Scripts 	scripts        `json:"scripts"`
 	Options     options        `json:"options"`
 
 	// The following are unused variables that are still silently accepted
 	// for compatibility with Heroku app.json files.
-
 	IgnoredDescription string   `json:"description"`
 	IgnoredKeywords    []string `json:"keywords"`
 	IgnoredLogo        string   `json:"logo"`
--- a/cmd/cloudshell_open/appfile.go
+++ b/cmd/cloudshell_open/appfile.go
@@ -40,7 +40,7 @@ type options struct {
 }
 
 type scripts struct {
-	Postdeploy string `json:"postdeploy"`
+	PostDeploy string `json:"postdeploy"`
 }
 
 type appFile struct {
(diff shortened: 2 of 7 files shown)


jamesward:
needs docs once we are good with the impl.

ahmetb:
What do you think of making the script options always a "file" (whose existence we check for at the beginning, so that it doesn't fail ahead of time).

This way, `"postdeploy": "cmd arg1 arg2 arg3"` maybe won't work, but we will allow executing `"postdeploy": "scripts/post-deploy.sh"`. 

Developers will need to make the file executable (+x) and we'd just exec them directly (without needing `sh -c`).

This sounds like a more bulletproof design to me.

cmd/cloudshell_open/appfile.go
```diff
@@ -32,9 +32,14 @@ type env struct {
 	Required    *bool  `json:"required"`
 }
 
+type scripts struct {
+	Postdeploy string `json:"postdeploy"`
```
ahmetb: I like this structure.

I'll now think if it makes sense if each member of `scripts` should be a list (`[]string`) so people can specify multiple script files.

ahmetb: PostDeploy is a better variable name in this case.

jamesward:
```suggestion
	PostDeploy string `json:"postdeploy"`
```

jamesward:
Initially I was trying to not deviate from the `app.json` spec, but I agree that it would be better to only support a list of filenames.  And since 100% portability of `app.json` files isn't a goal, we should do what makes the most sense for us.  So I'll update this.

jamesward:
This has been rebased on #88 so we should get that merged first.  Reason for that is that `postcreate` made more sense than `postdeploy` since the later put the challenge of not setting things up twice on the author.  Now we can check if the app is being created and if so, run `postcreate`.  In the future we can also add `precreate`, `postupdate`, etc.

Also, I changed it from using `scripts` since that would have made this incompatible with Heroku `app.json`.

README.md
```diff
@@ -45,17 +45,23 @@ For example:
...
+        },
+        "TITLE": {
+            "description": "title for your site"
+        },
+        "APP_SECRET": {
+            "generator": "secret"
+        }
+    },
+    "hook": {
+        "postcreate": "postcreate.sh"
```
ahmetb: # option1

- "./postcreate.sh --install" -> shlex (favored, for now)

- ["./postcreate.sh", "--install"]

execve($0)

# option2 

/bin/sh -c "$0"

```
"postcreate": "date\ninstall_foo\ninstall_bar" --> `;` vs `&&`
```

# single entrypoint

```
#!/usr/bin/env bash
set -euo pipefail

s=$FOO # "a b"
./bin/prep $s
./bin/setup "$secret"
```

```
  postcreate:
    - cmd: bin/prep
      args: [""]
    - cmd:
      args: []
```

## contract

1. env vars
2. cwd

## extensibility

```
{
  "hooks": {
     "postcreate": { <--- hook
        "command": "bin/setup"
        "commands": [
           "bin/setup",
        ]
        "timeout": 
     }
   }
}
```



```
{
  "hooks": {
     "postcreate": { <--- hook
        "commands": [
           "gsutil mb gs://$BUCKET",
        ]
        "timeout": 
     }
   }
}
```

os.Exec(cmd)

os.Exec("/bin/sh", "-c", "set -euo pipefail; "+cmd)

README.md
```diff
@@ -70,6 +76,9 @@ Reference:
   - `value`: _(optional)_ default value for the variable, should be a string.
   - `required`, _(optional, default: `true`)_ indicates if they user must provide
     a value for this variable.
+  - `generator`, _(optional)_ use a generator for the value, currently only support `secret`
+- `hook`: _(optional)_ Run scripts at phases of the deploy process
+  - `postcreate`: _(optional) Runs the specified script file after the service has been created
```
ahmetb: first-time create

README.md
```diff
@@ -70,6 +76,9 @@ Reference:
   - `value`: _(optional)_ default value for the variable, should be a string.
   - `required`, _(optional, default: `true`)_ indicates if they user must provide
     a value for this variable.
+  - `generator`, _(optional)_ use a generator for the value, currently only support `secret`
+- `hook`: _(optional)_ Run scripts at phases of the deploy process
```
ahmetb: + add pre-deploy:

cmd/cloudshell_open/scripts.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+)
+
+func runScript(dir, phase string, project string, service string, region string, command string) error {
+	cmd := exec.Command(command)
+	cmd.Env = []string{
```
ahmetb: just accept map[string]string

cmd/cloudshell_open/scripts.go
```diff
@@ -0,0 +1,38 @@
...
+		fmt.Sprintf("PROJECT_ID=%s", project),
+		fmt.Sprintf("SERVICE=%s", service),
+		fmt.Sprintf("REGION=%s", region)}
+	cmd.Dir = dir
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	cmd.Stdin = os.Stdin
```
ahmetb: fmt.Println("Executing command: %s")

cmd/cloudshell_open/scripts.go
```diff
@@ -0,0 +1,38 @@
...
+	err := cmd.Run()
+	if err != nil {
+		return fmt.Errorf("%s script failed: %v", phase, err)
+	}
```
ahmetb: move this out
this method should not know the phase.

cmd/cloudshell_open/scripts.go
```diff
@@ -0,0 +1,38 @@
...
+package main
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+)
+
+func runScript(dir, phase string, project string, service string, region string, command string) error {
+	cmd := exec.Command(command)
```
ahmetb:
```
if command has no space
  exec.Command(command)
else
  exec.Command("/bin/bash", "-c", "set -euo pipefail; "+command)
fi
```

cmd/cloudshell_open/scripts.go
```diff
@@ -0,0 +1,38 @@
...
+	cmd.Env = []string{
+		fmt.Sprintf("PROJECT_ID=%s", project),
+		fmt.Sprintf("SERVICE=%s", service),
+		fmt.Sprintf("REGION=%s", region)}
+	cmd.Dir = dir
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	cmd.Stdin = os.Stdin
+	err := cmd.Run()
```
ahmetb: return err directly

cmd/cloudshell_open/cloudrun.go
```diff
@@ -112,14 +188,28 @@ func tryFixServiceName(name string) string {
...
+	name = strings.ToLower(name)
+
+	reg := regexp.MustCompile("[^a-z0-9-]+")
+
+	name = reg.ReplaceAllString(name, "-")
+
+	if name[0] == '-' {
+		name = fmt.Sprintf("svc%s", name)
+	}
+
```
ahmetb: extract

jamesward:
I'm thinking through how we can have a `precreate` command that will create a Cloud SQL instance so that the service can startup on first revision with everything it needs.

We need to create a DB password which we can do with a generated secret env var.  Then use that password env var in `precreate` when we create the Cloud SQL instance.  But we also need to set an env var based on some values from `precreate` like: `CLOUD_SQL_CONNECTION_NAME=$project:$region:$cloudsqlinstance`
I'm not sure how to handle that part.

Another related one is something like:
```
{
  "hooks": {
    "precreate": {
      "commands": [
        "gsutil mb gs://$BUCKET"
      ]
    }
  }
}
```

We can get `BUCKET` from the env prompt but I think it needs to have the Project ID in it.  We could make the user enter it, but they've already selected it, so that seems weird.  It might be that we need to expand on the env var generator to enable this use case.

I don't think this requirement blocks this feature but it is something that we should think about.

jamesward:
Here is an example of the output with the `set -x`:
```
[ ! ]Running command:
	echo 'GONNA CREATE'
+ echo 'GONNA CREATE'
GONNA CREATE
[ ! ]Running command:
	date
+ date
Fri Jan 10 15:40:18 MST 2020
```
It is a bit verbose.

README.md
```diff
@@ -79,6 +91,11 @@ Reference:
   - `generator`, _(optional)_ use a generator for the value, currently only support `secret`
 - `options`: _(optional)_ Options when deploying the service
   - `allow-unauthenticated`: _(optional, default: `true`)_ allow unauthenticated requests
+- `hooks`: _(optional)_ Run scripts at phases of the deploy process
```
ahmetb: Can we add: 

> Specified commands run in separate bash shells with the environment variables configured for the application. Command outputs are shown as they are executed.

cmd/cloudshell_open/appfile_test.go
```diff
@@ -142,6 +142,14 @@ func Test_parseAppFile(t *testing.T) {
 						Required: &tru,
 					},
 				}}, false},
+		{"postcreate", `{
+			"hooks": {
+				"postcreate": {
```
ahmetb: let's add precreate as well, either combined to one test case or not.

cmd/cloudshell_open/main.go
```diff
@@ -308,6 +309,20 @@ func run(opts runOpts) error {
 		}
 	}
 
+	hookEnvs := append(envs,
+		fmt.Sprintf("PROJECT_ID=%s", project),
+		fmt.Sprintf("SERVICE=%s", serviceName),
+		fmt.Sprintf("REGION=%s", region))
+
```
ahmetb: let's also document that we add these.

ahmetb: hmm I wonder if we should consider making K_SERVICE instead of SERVICE, that's what https://cloud.google.com/run/docs/reference/container-contract does.

cmd/cloudshell_open/main.go
```diff
@@ -308,6 +309,20 @@ func run(opts runOpts) error {
...
+	if reflect.DeepEqual(existingService, service{}) {
```
ahmetb: ideally we should not use `reflect` in main code (tests are ok).

Let's just return `*service` from describeService (or whatever) so we can do `service == nil` check.

cmd/cloudshell_open/main.go
```diff
@@ -308,6 +309,20 @@ func run(opts runOpts) error {
...
+		for _, command := range appFile.Hooks.PreCreate.Commands {
+			err = runScript(appDir, command, hookEnvs)
+			if err != nil {
+				return err
```
ahmetb: let's format this error like this:

`fmt.Errorf("failed to execute precreate command[%d]: %v", i, err)`

keep in mind that `err` will most likely look like `exec: status 1` and not a lot more meaningful than that. error message never includes cmd's output (but we're printing it to the screen)

cmd/cloudshell_open/main.go
```diff
@@ -343,6 +358,15 @@ func run(opts runOpts) error {
...
 	}
 
+	if reflect.DeepEqual(existingService, service{}) {
+		for _, command := range appFile.Hooks.PostCreate.Commands {
+			err = runScript(appDir, command, hookEnvs)
+			if err != nil {
+				return err
+			}
+		}
+	}
```
ahmetb: can we refactor these hooks out, to something like


```
if err := executeHooks("postcreate", appFile.Hooks.PostCreate.Commands); err != nil { return err }
```

you can construct the detailed error message in `executeHooks`.

ahmetb: this would also

- simplify/shorten this method
- prevent duplicating this logic (e.g. for loops around runScript)

cmd/cloudshell_open/main.go
```diff
@@ -308,6 +309,20 @@ func run(opts runOpts) error {
 		}
 	}
 
+	hookEnvs := append(envs,
```
ahmetb: so there's a bug here:

if you do `append(envs, something...)`, it actually overwrites `env` slice as well. the underlying slice is modified.

I recommend doing this instead

```
foo = append([]string{v1, v2, v3}, envs...)
```

cmd/cloudshell_open/scripts.go
```diff
@@ -0,0 +1,38 @@
...
+	"os"
+	"os/exec"
+
+	"github.com/fatih/color"
+)
+
+var cmdColor = color.New(color.FgHiBlue)
+
+func runScript(dir, command string, envs []string) error {
+	fmt.Println(infoPrefix + "Running command:")
```
ahmetb: we can make this more compact, like:

    infoPrefix + "Running command: " + color.GrayString(cmd)

all on the same line.

cmd/cloudshell_open/scripts.go
```diff
@@ -0,0 +1,38 @@
...
+	cmdColor.Printf("\t%s\n", command)
+
+	cmd := exec.Command("/bin/bash", "-c", "set -euo pipefail; set -x; " + command)
+	cmd.Env = envs
+	cmd.Dir = dir
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	cmd.Stdin = os.Stdin
```
ahmetb: Idea:

we can replace os.Stdout and os.Stderr with `color.New(color.FgGray)` and pass here (because, *color.Color is an `io.Writer`).

That way, all cmd outputs will be grayed out. Food for thought.

ahmetb: if that doesn't work, something like this should:


```
type mywriter struct { out io.Writer }

func (m myWriter) Write(p []byte) (int, error) {
    return color.New(color.FgGray).Fprintf(m.out, p)
}

use os.Stdout = myWriter{os.Stderr}
```

jamesward: I like it, but I'm probably not understanding something with Go, but here is what I get:

```
Cannot use 'color.New(color.FgBlue)' (type *Color) as type io.Writer in assignment Type does not implement 'io.Writer' as some methods are missing: Write(p []byte) (n int, err error) 
```

ahmetb: Yeah the 2nd example I gave above can solve it

jamesward: I've updated this.  Let me know how it looks.

ahmetb:
ok if you've verified it works we can merge.

README.md
```diff
@@ -79,6 +91,11 @@ Reference:
   - `generator`, _(optional)_ use a generator for the value, currently only support `secret`
 - `options`: _(optional)_ Options when deploying the service
   - `allow-unauthenticated`: _(optional, default: `true`)_ allow unauthenticated requests
+- `hooks`: _(optional)_ Run commands in separate bash shells with the environment variables configured for the application and environment variables for `PROJECT_ID`, `K_SERVICE`, and `REGION`. Command outputs are shown as they are executed.
```
ahmetb: just a final thought here:

- I think if we be more specific, like GOOGLE_CLOUD_PROJECT and GOOGLE_CLOUD_REGION, that would be better. (as a matter of fact many of our client libs support GOOGLE_CLOUD_PROJECT ootb)

- please do 80-char line wraps (use vscode + markdownlint + rewrap)

jamesward:
Verified to work with: https://github.com/jamesward/nodebars/tree/buttonify

cmd/cloudshell_open/scripts.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func runScript(dir, command string, envs []string) error {
+	fmt.Println(infoPrefix + " Running command: " + color.BlueString(command))
+
+	cmd := exec.Command("/bin/bash", "-c", "set -euo pipefail; set -x; " + command)
+	cmd.Env = envs
+	cmd.Dir = dir
+	cmd.Stdout = myWriter{os.Stdout, color.FgGreen}
+	cmd.Stderr = myWriter{os.Stderr, color.FgRed}
```
ahmetb: I think stderr==red rarely works out for good. It scares users much more than it helps, even curl prints progress to stderr, and we don't want all that to be red. You can fix later.

It would be better if these were just gray (FgHiBlack)

ahmetb:
Good to go 👍

cmd/cloudshell_open/scripts.go
jamesward: That makes sense.  Thanks!

jamesward:
Thanks for all your help!

PR #88, opened 2019-09-24
add env secret generator

jamesward:
Tested with https://github.com/jamesward/cloud-run-button-secret-test

### DIFF
--- a/cmd/cloudshell_open/appfile.go
+++ b/cmd/cloudshell_open/appfile.go
@@ -15,11 +15,13 @@
 package main
 
 import (
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"io"
 	"os"
 	"path/filepath"
+	"crypto/rand"
 
 	"github.com/fatih/color"
 
@@ -30,6 +32,7 @@ type env struct {
 	Description string `json:"description"`
 	Value       string `json:"value"`
 	Required    *bool  `json:"required"`
+	Generator   string `json:"generator"`
 }
 
 type options struct {
@@ -109,6 +112,26 @@ func getAppFile(dir string) (appFile, error) {
 	return *af, nil
 }
 
+func rand64String() (string, error) {
+	b := make([]byte, 64)
+	if _, err := rand.Read(b); err != nil {
+		return "", err
+	}
+	return base64.StdEncoding.EncodeToString(b), nil
+}
+
+// takes the envs defined in app.json, and the existing envs and returns the new envs that need to be prompted for
+func prepEnv(list map[string]env, existing map[string]struct{}) map[string]env {
+	for k := range list {
+		_, isPresent := existing[k]
+		if isPresent {
+			delete(list, k)
+		}
+	}
+
+	return list
+}
+
 func promptEnv(list map[string]env) ([]string, error) {
 	// TODO(ahmetb): remove these defers and make customizations at the
 	// individual prompt-level once survey lib allows non-global settings.
@@ -118,19 +141,30 @@ func promptEnv(list map[string]env) ([]string, error) {
 	// field and prompt the questions as they appear in the app.json file as
 	// opposed to random order we do here.
 	for k, e := range list {
-		var resp string
-		if err := survey.AskOne(&survey.Input{
-			Message: fmt.Sprintf("Value of %s environment variable (%s)",
-				color.CyanString(k),
-				color.HiBlackString(e.Description)),
-			Default: e.Value,
-		}, &resp,
-			survey.WithValidator(survey.Required),
-			surveyIconOpts,
-		); err != nil {
-			return nil, fmt.Errorf("failed to get a response for environment variable %s", k)
+
+		if e.Generator == "secret" {
+			resp, err := rand64String()
+			if err != nil {
+				return nil, fmt.Errorf("failed to generate secret for %s - %v", k, err)
+			}
+			out = append(out, k+"="+resp)
+		} else {
+			var resp string
+
+			if err := survey.AskOne(&survey.Input{
+				Message: fmt.Sprintf("Value of %s environment variable (%s)",
+					color.CyanString(k),
+					color.HiBlackString(e.Description)),
+				Default: e.Value,
+			}, &resp,
+				survey.WithValidator(survey.Required),
+				surveyIconOpts,
+			); err != nil {
+				return nil, fmt.Errorf("failed to get a response for environment variable %s", k)
+			}
+			out = append(out, k+"="+resp)
 		}
-		out = append(out, k+"="+resp)
+
 	}
 	return out, nil
 }
--- a/cmd/cloudshell_open/appfile.go
+++ b/cmd/cloudshell_open/appfile.go
@@ -145,7 +145,7 @@ func promptEnv(list map[string]env) ([]string, error) {
 		if e.Generator == "secret" {
 			resp, err := rand64String()
 			if err != nil {
-				return nil, fmt.Errorf("failed to generate secret for %s - %v", k, err)
+				return nil, fmt.Errorf("failed to generate secret for %s : %v", k, err)
 			}
 			out = append(out, k+"="+resp)
 		} else {
--- a/cmd/cloudshell_open/appfile.go
+++ b/cmd/cloudshell_open/appfile.go
@@ -86,6 +86,12 @@ func parseAppFile(r io.Reader) (*appFile, error) {
 		v.Env[k] = env
 	}
 
+	for k, env := range v.Env {
+		if env.Generator == "secret" && env.Value != "" {
+			return nil, fmt.Errorf("Env Var '%s' can't have both a value and use the secret generator", k)
+		}
+	}
+
 	return &v, nil
 }
 
--- a/cmd/cloudshell_open/appfile.go
+++ b/cmd/cloudshell_open/appfile.go
@@ -88,7 +88,7 @@ func parseAppFile(r io.Reader) (*appFile, error) {
 
 	for k, env := range v.Env {
 		if env.Generator == "secret" && env.Value != "" {
-			return nil, fmt.Errorf("Env Var '%s' can't have both a value and use the secret generator", k)
+			return nil, fmt.Errorf("env var %q can't have both a value and use the secret generator", k)
 		}
 	}
 
@@ -127,7 +127,7 @@ func rand64String() (string, error) {
 }
 
 // takes the envs defined in app.json, and the existing envs and returns the new envs that need to be prompted for
-func prepEnv(list map[string]env, existing map[string]struct{}) map[string]env {
+func needEnvs(list map[string]env, existing map[string]struct{}) map[string]env {
 	for k := range list {
 		_, isPresent := existing[k]
 		if isPresent {
@@ -136,6 +136,44 @@ func prepEnv(list map[string]env, existing map[string]struct{}) map[string]env {
 	}
 
 	return list
+}
+
+func promptOrGenerateEnvs(list map[string]env) ([]string, error) {
+	var toGenerate []string
+	var toPrompt = make(map[string]env)
+
+	for k, e := range list {
+		if e.Generator == "secret" {
+			toGenerate = append(toGenerate, k)
+		} else {
+			toPrompt[k] = e
+		}
+	}
+
+	generated, err := generateEnvs(toGenerate)
+	if err != nil {
+		return nil, err
+	}
+
+	prompted, err := promptEnv(toPrompt)
+	if err != nil {
+		return nil, err
+	}
+
+	return append(generated, prompted...), nil
+}
+
+func generateEnvs(keys []string) ([]string, error) {
+
+	for i, key := range keys {
+		resp, err := rand64String()
+		if err != nil {
+			return nil, fmt.Errorf("failed to generate secret for %s : %v", key, err)
+		}
+		keys[i] = key+"="+resp
+	}
+
+	return keys, nil
 }
 
 func promptEnv(list map[string]env) ([]string, error) {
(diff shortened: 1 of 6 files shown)


jamesward:
needs docs once we are good with the impl

ahmetb:
The update case has been bothering me: What if I deploy first-time and generate a `password1` with this. Then, I provision a bucket or a DB with this `password1`.

Now I re-deploy by clicking the button, and I will get `password2`. This will cause either a conflict (like invalidated cookies), or a new DB/storage to be provisioned if I was relying on this value to stay the same between re-deploys.

How can we achieve this? What's your experience about this from prior art?

jamesward:
Yeah, good catch.  We shouldn't overwrite these.  So I'll add a check for that.

ahmetb:
I'm also not sure how to check for that. We currently don't read if the service exists first. Maybe instead of --set-env-vars we can use --env-vars but I am not certain if it will do what we want here.

Also it might pose a discrepancy between the command we show vs what really happens.

Can we hold off until this change for a while perhaps and see what some use cases look like?

ahmetb:
Putting this on `hold` until we figure out a cleaner design on how we can make this work reproducibly (i.e. without destroying the old value, if exists).

If anyone needs this feature, please add comments.

jamesward:
@ahmetb I've updated this to not overwrite existing generated env vars. LMKWYT

cmd/cloudshell_open/cloudrun.go
```diff
@@ -87,12 +87,14 @@ func promptDeploymentRegion(ctx context.Context, project string) (string, error)
...
 
-func serviceURL(project, name, region string) (string, error) {
+
+func describe(project, name, region, format, flatten string) (string, error) {
 	cmd := exec.Command("gcloud", "run", "services", "describe", name,
 		"--project", project,
 		"--platform", "managed",
 		"--region", region,
-		"--format", "value(status.address.url)")
+		"--format", format,
```
ahmetb: IMO let's just do `--format=json` here and parse it into a legit Service object and return it. This is opening a large can of worms (i.e. Test_describe) and complicated `--format` strings.

cmd/cloudshell_open/cloudrun.go
```diff
@@ -40,7 +40,7 @@ func deploy(project, name, image, region string, envs []string) (string, error)
 		"--region", region,
 		"--memory", defaultRunMemory,
 		"--allow-unauthenticated",
-		"--set-env-vars", strings.Join(envs, ","))
+		"--update-env-vars", strings.Join(envs, ","))
```
ahmetb: Do we need to change this, since we're omitting the env var for secret?
Does --set-env-vars actually clear all + set specified?

cmd/cloudshell_open/appfile.go
```diff
@@ -113,19 +116,35 @@ func promptEnv(list map[string]env) ([]string, error) {
...
-			Default: e.Value,
-		}, &resp,
-			survey.WithValidator(survey.Required),
-			surveyIconOpts,
-		); err != nil {
-			return nil, fmt.Errorf("failed to get a response for environment variable %s", k)
+		_, isPresent := existing[k]
+
+		if e.Generator == "secret" {
+			if !isPresent {
```
ahmetb: it would be much better if you do `if isPresent { continue }` than nesting it.

cmd/cloudshell_open/appfile.go
```diff
@@ -113,19 +116,35 @@ func promptEnv(list map[string]env) ([]string, error) {
...
-			Message: fmt.Sprintf("Value of %s environment variable (%s)",
-				color.CyanString(k),
-				color.HiBlackString(e.Description)),
-			Default: e.Value,
-		}, &resp,
-			survey.WithValidator(survey.Required),
-			surveyIconOpts,
-		); err != nil {
-			return nil, fmt.Errorf("failed to get a response for environment variable %s", k)
+		_, isPresent := existing[k]
```
ahmetb: I think `promptEnv` (by its name) "check for existing stuff and generate "secret" env vars" is the wrong place to do it

Let's create a new `prepEnv` method that does the changes you added below, and calls into `promptEnv` actually for prompting purposes only.

cmd/cloudshell_open/cloudrun.go
jamesward: Yeah, that makes sense.  I think I'll just have to create a bunch of structs to hold the fairly large tree.  Or can I use Knative as a library?

cmd/cloudshell_open/appfile.go
```diff
@@ -113,19 +116,35 @@ func promptEnv(list map[string]env) ([]string, error) {
...
+
+		if e.Generator == "secret" {
+			if !isPresent {
+				b := make([]byte, 64)
+				_, err := rand.Read(b)
+				if err != nil {
+					return nil, fmt.Errorf("failed to generate secret for %s - %v", k, err)
+				}
+				resp := base64.StdEncoding.EncodeToString(b)
```
ahmetb: you can make this a new small method.

cmd/cloudshell_open/cloudrun.go
jamesward: Yeah, `--set-env-vars` clears and sets.

cmd/cloudshell_open/cloudrun.go
ahmetb: I think we can very simply create a tiny new struct that has fields for

- spec.containers[].env[]
- status.address.url (or whatever, I forgot)

Importing Knative struct here isn't gonna do much good tbh, since we don't use most of it. But it'll increase the binary size significantly since it imports all Kubernetes API machinery code.

Maybe the cloudrun api clients we have actually have this type.

In that case, maybe we can even eliminate call to `gcloud run services describe` with a proper GetService API call.

jamesward: Yeah, it would be nice to use the client APIs.  I can't find a way for `gcloud` `format=json` to give me a child, like:
```
gcloud run services describe hello-netcat --platform=managed --project=jw-demo --region=us-central1 --format="json(spec.template)"
{
  "spec": {
    "template": {
      "metadata": {
        "annotations": {
          "client.knative.dev/user-image": "gcr.io/jw-demo/hello-netcat",
          "run.googleapis.com/client-name": "gcloud",
          "run.googleapis.com/client-version": "272.0.0"
        },
        "name": "hello-netcat-00003-vaq"
      },
      "spec": {
        "containerConcurrency": 0,
        "containers": [
          {
            "image": "gcr.io/jw-demo/hello-netcat",
            "resources": {
              "limits": {
                "memory": "512Mi"
              }
            }
          }
        ]
      }
    }
  }
}
```

jamesward: e.g. `format="json(spec.template)"` reduces the tree but doesn't crop it.

jamesward:
This has been refactored.  Really appreciate the code reviews!

cmd/cloudshell_open/appfile.go
```diff
@@ -118,19 +141,30 @@ func promptEnv(list map[string]env) ([]string, error) {
...
-				color.CyanString(k),
-				color.HiBlackString(e.Description)),
-			Default: e.Value,
-		}, &resp,
-			survey.WithValidator(survey.Required),
-			surveyIconOpts,
-		); err != nil {
-			return nil, fmt.Errorf("failed to get a response for environment variable %s", k)
+
+		if e.Generator == "secret" {
```
ahmetb: where's the validation that validates that `generator` and `value` cannot be set at the same time?

cmd/cloudshell_open/cloudrun.go
```diff
@@ -32,10 +34,44 @@ const (
...
+}
+
+type serviceStatus struct {
+	Address address `json:"address"`
+}
+
+type address struct {
+	Url string `json:"url"`
+}
+
```
ahmetb: you don't need these, just do it like this (inline anonymous type)


```
var v struct {
    Spec struct{
      Template struct { 
      } `json:"template"` 
    } `json:"spec"`
}
```

in the method that does unmarshal

ahmetb: actually ignore the "inline" part.

it seems like you're not just using Url field. (but also env)
in that case, you can still create 1 type called Service, and kinda inline everything else into that type def.

cmd/cloudshell_open/cloudrun.go
```diff
@@ -102,18 +139,57 @@ func promptDeploymentRegion(ctx context.Context, project string) (string, error)
...
-	return strings.TrimSpace(string(b)), nil
+
+	return service.Status.Address.Url, nil
+}
+
+func envVars(project, name, region string) (map[string]struct{}, error) {
+	service, err := describe(project, name, region)
+
+	if err != nil {
+		return map[string]struct{}{}, err
```
ahmetb: you can do `return nil, err`

cmd/cloudshell_open/cloudrun.go
```diff
@@ -102,18 +139,57 @@ func promptDeploymentRegion(ctx context.Context, project string) (string, error)
...
+	}
+
+	existing := map[string]struct{}{}
```
ahmetb: use `make(map[..]..)`

cmd/cloudshell_open/cloudrun_test.go
```diff
@@ -50,3 +54,52 @@ func Test_tryFixServiceName(t *testing.T) {
 		})
 	}
 }
+
+// this integration test depends on gcloud being auth'd as someone that has access to the crb-test project
```
ahmetb: let's delete this test

cmd/cloudshell_open/cloudrun_test.go
```diff
@@ -50,3 +54,52 @@ func Test_tryFixServiceName(t *testing.T) {
...
+			}
+
+			if err != nil && !tt.wantErr {
+				t.Error(tparams + " produced an error: %s", err)
+			}
+		})
+	}
+}
+
+// this integration test depends on gcloud being auth'd as someone that has access to the crb-test project
```
ahmetb: let's delete this test as well
if you want to really test parsing, I suggest we add testdata/example-svc.yaml and we can parse that.

cmd/cloudshell_open/main.go
```diff
@@ -242,6 +238,20 @@ func run(opts runOpts) error {
 
 	image := fmt.Sprintf("gcr.io/%s/%s", project, serviceName)
 
+	existingEnvVars := map[string]struct{}{}
+	_, err = describe(project, serviceName, region)
+	if err == nil {
```
ahmetb: we're not checking actual error here.
so it seems like we actually need something along the lines of checking service exists `(bool, error)` instead of just ignoring `err` here.

this is hacky currently, but it'll work, but it might mask other issues in the future.

cmd/cloudshell_open/main.go
```diff
@@ -312,11 +322,14 @@ func run(opts runOpts) error {
 	cmdColor.Printf("\t  --image=%s", parameter(image))
 	cmdColor.Println("\\")
 	cmdColor.Printf("\t  --memory=%s", parameter(defaultRunMemory))
+
 	for _, optionFlag := range optionsFlags {
 		cmdColor.Println("\\")
 		cmdColor.Printf("\t  %s\n", optionFlag)
 	}
 
+	// todo(jamesward) display env vars flag
```
ahmetb: address todo

cmd/cloudshell_open/cloudrun.go
```diff
@@ -32,10 +34,44 @@ const (
...
+type address struct {
+	Url string `json:"url"`
+}
+
 func optionsToFlags(options options) []string {
-	authSetting := "--no-allow-unauthenticated"
-	if *options.AllowUnauthenticated {
-		authSetting = "--allow-unauthenticated"
+	authSetting := "--allow-unauthenticated"
+	if options.AllowUnauthenticated != nil && *options.AllowUnauthenticated == false {
```
ahmetb: I think the PRs got mixed together. :) maybe a rebase will help.

cmd/cloudshell_open/appfile.go
```diff
@@ -118,19 +141,30 @@ func promptEnv(list map[string]env) ([]string, error) {
...
+			resp, err := rand64String()
+			if err != nil {
+				return nil, fmt.Errorf("failed to generate secret for %s - %v", k, err)
```
ahmetb: let's replace ` - ` with `: .`

jamesward:
```suggestion
				return nil, fmt.Errorf("failed to generate secret for %s : %v", k, err)
```

cmd/cloudshell_open/cloudrun.go
jamesward: Moved out to another PR: https://github.com/GoogleCloudPlatform/cloud-run-button/pull/127

cmd/cloudshell_open/cloudrun.go
jamesward:
```suggestion
		return nil, err
```

cmd/cloudshell_open/main.go
jamesward: Yeah, the error could be anything.  We don't have a good way when wrapping the `gcloud` command to parse actual errors.  But we could use the API directly.  I'll add a todo for now.

cmd/cloudshell_open/appfile.go
jamesward: Added in the parsing step

cmd/cloudshell_open/cloudrun.go
```diff
@@ -102,18 +125,57 @@ func promptDeploymentRegion(ctx context.Context, project string) (string, error)
...
+func envVars(project, name, region string) (map[string]struct{}, error) {
+	service, err := describe(project, name, region)
 
-	b, err := cmd.CombinedOutput()
 	if err != nil {
-		return "", fmt.Errorf("deployment to Cloud Run failed: %+v. output:\n%s", err, string(b))
+		return nil, err
+	}
+
+	existing := make(map[string]struct{})
```
ahmetb: map[string]bool

cmd/cloudshell_open/appfile.go
```diff
@@ -83,6 +86,12 @@ func parseAppFile(r io.Reader) (*appFile, error) {
 		v.Env[k] = env
 	}
 
+	for k, env := range v.Env {
+		if env.Generator == "secret" && env.Value != "" {
+			return nil, fmt.Errorf("Env Var '%s' can't have both a value and use the secret generator", k)
```
ahmetb: Env -> env

ahmetb: '%s' --> %q

cmd/cloudshell_open/appfile.go
```diff
@@ -118,19 +147,30 @@ func promptEnv(list map[string]env) ([]string, error) {
...
-				color.CyanString(k),
-				color.HiBlackString(e.Description)),
-			Default: e.Value,
-		}, &resp,
-			survey.WithValidator(survey.Required),
-			surveyIconOpts,
-		); err != nil {
-			return nil, fmt.Errorf("failed to get a response for environment variable %s", k)
+
+		if e.Generator == "secret" {
```
ahmetb: move out of promptEnv

cmd/cloudshell_open/cloudrun.go
jamesward: It appears that `map[string]struct{}` is the connonical way to do a set in Go.  But I'm not certain. Let me know if you'd rather do it via `map[string]bool`

ahmetb: yeah it's all the same since you use the keys as a set.

PR #89, opened 2019-09-24
add require-auth setting - fixes #83

### LINKED ISSUE #83: app.json support for "Require authentication"
I'm working on an app that'd be nice to buttonize but it is one that should set the "Require authentication" flag on the service.  I think if we allow the `app.json` to have a property like:
```
{
    "require-auth": true
}
```
Then maybe prompt for the roles to allow.

This might be related to #33 but I'm not totally sure.

jamesward:
Tested with: https://github.com/jamesward/cloud-run-button-require-auth-test

### DIFF
--- a/cmd/cloudshell_open/appfile.go
+++ b/cmd/cloudshell_open/appfile.go
@@ -33,8 +33,9 @@ type env struct {
 }
 
 type appFile struct {
-	Name string         `json:"name"`
-	Env  map[string]env `json:"env"`
+	Name        string         `json:"name"`
+	Env         map[string]env `json:"env"`
+	RequireAuth *bool          `json:"require-auth"`
 
 	// The following are unused variables that are still silently accepted
 	// for compatibility with Heroku app.json files.
@@ -76,6 +77,11 @@ func parseAppFile(r io.Reader) (*appFile, error) {
 			env.Required = &v
 		}
 		v.Env[k] = env
+	}
+
+	// make "require-auth" false by default
+	if v.RequireAuth == nil {
+		v.RequireAuth = new(bool)
 	}
 
 	return &v, nil
--- a/cmd/cloudshell_open/appfile.go
+++ b/cmd/cloudshell_open/appfile.go
@@ -32,10 +32,14 @@ type env struct {
 	Required    *bool  `json:"required"`
 }
 
+type options struct {
+	AllowUnauthenticated *bool `json:"allow-unauthenticated"`
+}
+
 type appFile struct {
 	Name        string         `json:"name"`
 	Env         map[string]env `json:"env"`
-	RequireAuth *bool          `json:"require-auth"`
+	Options     options        `json:"options"`
 
 	// The following are unused variables that are still silently accepted
 	// for compatibility with Heroku app.json files.
@@ -77,11 +81,6 @@ func parseAppFile(r io.Reader) (*appFile, error) {
 			env.Required = &v
 		}
 		v.Env[k] = env
-	}
-
-	// make "require-auth" false by default
-	if v.RequireAuth == nil {
-		v.RequireAuth = new(bool)
 	}
 
 	return &v, nil
--- a/cmd/cloudshell_open/appfile_test.go
+++ b/cmd/cloudshell_open/appfile_test.go
@@ -88,13 +88,14 @@ func Test_parseAppFile(t *testing.T) {
 		wantErr bool
 	}{
 		{"empty json is EOF", ``, nil, true},
-		{"empty object ok", `{}`, &appFile{}, false},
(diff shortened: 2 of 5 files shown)


jamesward:
needs docs, once we are good with the impl.

ahmetb:
We might be better off extending the app.json more systemically and carefully. 

Also, soon enough we'll be asked to add options for cpu, memory, serviceAccount, ... Those are part of the `spec`, but making an app public is actually not a part of Service object (it's an IAM role binding).

That's why I'm not positive about a top-level `requireAuth` field. Something like this could be better:

```
{
  ...

  "authentication": {
     "allowUsers": [
        "allUsers"
     ]
  },

  "spec":  { ... }
}
```

wdyt?

cmd/cloudshell_open/appfile.go
```diff
@@ -33,8 +33,9 @@ type env struct {
 }
 
 type appFile struct {
-	Name string         `json:"name"`
-	Env  map[string]env `json:"env"`
+	Name        string         `json:"name"`
+	Env         map[string]env `json:"env"`
+	RequireAuth *bool          `json:"require-auth"`
```
ahmetb: You can drop `*` and just make it false-by-default.

Since we don't serialize this object, having the extra field listed doesn't make a difference.

It would be useful to have `*bool` when/if you're trying to distinguish `nil` (unset) vs `false` (zero value).

cmd/cloudshell_open/appfile.go
```diff
@@ -78,6 +79,11 @@ func parseAppFile(r io.Reader) (*appFile, error) {
 		v.Env[k] = env
 	}
 
+	// make "require-auth" false by default
+	if v.RequireAuth == nil {
+		v.RequireAuth = new(bool)
+	}
```
ahmetb: ditto: using `bool` above eliminates this block

cmd/cloudshell_open/appfile_test.go
```diff
@@ -88,25 +88,26 @@ func Test_parseAppFile(t *testing.T) {
 		wantErr bool
 	}{
 		{"empty json is EOF", ``, nil, true},
-		{"empty object ok", `{}`, &appFile{}, false},
+		{"empty object ok", `{}`, &appFile{RequireAuth: &fals}, false},
```
ahmetb: ditto: using `bool` above solves specifying these values everywhere in the tests.

cmd/cloudshell_open/main.go
```diff
@@ -260,6 +260,11 @@ func run(c *cli.Context) error {
 		return fmt.Errorf("failed to push image to %s: %+v", image, err)
 	}
 
+	requireAuth := false
+	if appFile.RequireAuth != nil {
+		requireAuth = *appFile.RequireAuth
+	}
```
ahmetb: ditto: using `bool` above eliminates this block

cmd/cloudshell_open/main.go
```diff
@@ -274,12 +279,16 @@ func run(c *cli.Context) error {
...
+		cmdColor.Printf("\t  --no-allow-unauthenticated\n")
+	} else {
+		cmdColor.Printf("\t  --allow-unauthenticated\n")
+	}
 
 	end = logProgress(fmt.Sprintf("Deploying service %s to Cloud Run...", serviceLabel),
 		fmt.Sprintf("Successfully deployed service %s to Cloud Run.", serviceLabel),
 		"Failed deploying the application to Cloud Run.")
-	url, err := deploy(project, serviceName, image, region, envs)
+	url, err := deploy(project, serviceName, image, region, envs, requireAuth)
```
ahmetb: aside: I feel like we'll soon have to introduce a DeployOptions property-bag here :)

jamesward:
Been thinking about the `app.json` schema for this a bit more and I think we should actually try to match the CLI flags more closely.  So something like:
```
{
    ...
    "options": {
        "concurrency": 80,
        "memory": "512M",
        "no-allow-unauthenticated": true
    }
}
```

Since the default is to include `--allow-unauthenticated` we wouldn't need to have a setting for that unless we also wanted to support `"allow-unauthenticated": false` as another way to change the default.  WDYT?

jamesward:
I've updated this to include a bag of options, right now only `allow-unauthenticated`.

ahmetb:
I've been debating myself about `allowUnauthenticated` vs `allow-unauthenticated`. as the field name in app.json. I prefer earlier (and our API conventions do so as well). But I think either is OK, since we're manufacturing this new format.

Feel free to merge at your own pace. 🎄 🎅

jamesward:
Or I think more typical JSON would be `allow_unauthenticated` but I'm fine with whatever.  I'm debating if the root key should be `gcloud` instead of `options`.

PR #90, opened 2019-09-25
Increase caching and reduce layers in Dockerfile

### LINKED ISSUE #93: Use pack CLI in Cloud Shell
Cloud Shell will soon ship the most recent `pack` CLI (v0.4.1) for Buildpacks. There have been many changes to the Buildpacks spec; hopefully nothing breaks.

Also, the binary will be placed under `/usr/local/bin/pack` where the current `Dockerfile` tries to create a symlink. The symlink command will probably break by then.

chanseokoh:
UPDATE: `pack` is now included in Cloud Shell. No need to download.

Now, whenever I modify any file in the repo (the most frequent changes), the Docker has to re-download the pack when building. Since these are far less frequently updated, ~~I'd like to move up the layers.~~ Fixes #93. 

Also, I tried to reduce the number of layers in the final image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,5 @@
 FROM golang:1.12-alpine AS build
 RUN apk add --no-cache git
-WORKDIR /tmp
-RUN wget https://github.com/buildpack/pack/releases/download/v0.3.0/pack-v0.3.0-linux.tgz
-RUN tar -xzf pack-v0.3.0-linux.tgz
-RUN wget https://raw.githubusercontent.com/buildpack/pack/v0.3.0/LICENSE
 WORKDIR /src
 COPY go.mod go.sum ./
 RUN go mod download
@@ -12,7 +8,5 @@ RUN CGO_ENABLED=0 GOOS=linux go build -a -ldflags '-extldflags "-static"' \
         -o /bin/a.out ./cmd/cloudshell_open
 
 FROM gcr.io/cloudshell-images/cloudshell:latest
+RUN rm /google/devshell/bashrc.google.d/cloudshell_open.sh
 COPY --from=build /bin/a.out /bin/cloudshell_open
-COPY --from=build /tmp/pack /tmp/LICENSE /opt/pack/
-RUN rm /google/devshell/bashrc.google.d/cloudshell_open.sh && \
-        ln -s /opt/pack/pack /usr/local/bin/pack
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,19 +1,18 @@
 FROM golang:1.12-alpine AS build
 RUN apk add --no-cache git
+WORKDIR /tmp
+RUN wget https://github.com/buildpack/pack/releases/download/v0.3.0/pack-v0.3.0-linux.tgz
+RUN tar -xzf pack-v0.3.0-linux.tgz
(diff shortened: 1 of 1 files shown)


jamesward:
Nice!  This is great.  I'll wait for someone to give a thumbs up who better understands docker than I do.  But I really appreciate this contribution!

ahmetb:
There’s an even better approach: create a build stage for pack and copy it to the final stage from there :)

But this really shouldn't matter because ~nobody builds this image manually. The Dockerfile isnt really for developers; it’s to release a custom cloudshell image.

chanseokoh:
> The Dockerfile isnt really for developers; it’s to release a custom cloudshell image.

The [CONTRIBUTING](https://github.com/GoogleCloudPlatform/cloud-run-button/blob/master/CONTRIBUTING.md#test-cloud-run-buttons-underlying-command-locally) guide says to use this Dockerfile for local development, so I'm using it. Is there another way for me to build and test changes?

ahmetb:
🤔 we should probably consider updating that. 

Building the dockerfile locally will take quite long due to the base image size.

 Small changes can be tested by just compiling the program and running it directly like Cloud Shell invokes it (can be seen when the button is used). At least thats how I developed this repo. 🙃 It works also because gcloud is already authenticated.

chanseokoh:
Well, in any case, I think this is an improvement. And I don't really think using a separate build stage is necessarily a better idea, since you'll create another intermediate image in your local Docker. And if you use a different image than `golang:1.12-alpine`, you'll have to download that too. And if you use `golang:1.12-alpine`, then I think there's no advantage really to have a separate stage just for this tiny thing.

jamesward:
I also don't really know how to do much with Go, so have been using the Docker method.  I added #92 to track adding docs on local dev.

As far as this PR goes, I really don't know enough about docker to help much.  Except to point out that any changes to our docker image will affect the Cloud Shell launch process and could potentially speed it up or slow it down.

Dockerfile
```diff
@@ -1,19 +1,18 @@
 FROM golang:1.12-alpine AS build
 RUN apk add --no-cache git
+WORKDIR /tmp
+RUN wget https://github.com/buildpack/pack/releases/download/v0.3.0/pack-v0.3.0-linux.tgz
+RUN tar -xzf pack-v0.3.0-linux.tgz
+RUN wget https://raw.githubusercontent.com/buildpack/pack/v0.3.0/LICENSE
```
geshan: These 3 `RUN` can also be packed as you did below, something like:

```
RUN wget https://github.com/buildpack/pack/releases/download/v0.3.0/pack-v0.3.0-linux.tgz && \
       tar -xzf pack-v0.3.0-linux.tgz && \
       wget https://raw.githubusercontent.com/buildpack/pack/v0.3.0/LICENSE
```

chanseokoh: Yeah, that works, but since this is an intermediate image, we don't have to squash layers too aggressively. I think maximizing cache potential is better in this case.

chanseokoh:
Updated to stop installing pack. Fixes #93. Closes #87.

jamesward:
Does this mean it's in our base image now?

jamesward:
Ah, just saw: https://github.com/GoogleCloudPlatform/cloud-run-button/issues/93#issuecomment-538569692

Issue #85, opened 2019-09-19
Build Script to Enable Jib and Other Builds


jamesward:
It'd be nice if we could enable a way to run Jib builds via Maven or Gradle.  The `app.json` could include something like:
```
{
    "scripts": {
        "build": "./mvnw compile jib:build -Dimage=gcr.io/${PROJECT_ID}/hello-springboot"
    }
}
```
If that is set, we run the command instead of `docker build` or buildpack stuff.  We'd have to set the `PROJECT_ID` env var when we run the build command.

Note that this example requires a JVM which Cloud Shell has, but if the user needs other system deps they could add a build shell script to install them, and then specify that script.

ahmetb:
What if we brought the cloudbuild support back? That would eliminate the need for this.

Otherwise a more extensible design would be preferable that allows us plugging in cmd, docker, cloudbuild.

jamesward:
Yeah, I like that better than adding yet another way to run the build.  For reference, the feature request for Cloud Build support is: https://github.com/GoogleCloudPlatform/cloud-run-button/issues/14

ahmetb:
Honestly I feel like this request is valid, because I can see people would do non-trivial multi-step builds.

That's where arbitrary commands like Jib or ko can help, but granted, we need to bring them to the build environment by pre-installing.

jamesward:
Are you saying we should do something like above in addition to Cloud Build support?

jamesward:
I'm thinking about adding this since we've already figured out how to add / run scripts.  So we could allow a custom build via:
```
    "hooks": {
        "build": {
            "commands": [
                "./foo.sh"
            ]
        }
    }
```

This would take precedence over `Dockerfile`, Maven Jib, and Buildpacks.  Adding a 4th way to build isn't ideal, but I continue to run across places where my build produces the container image and there isn't another good way to handle this (except maybe Cloud Build support).

PR #91, opened 2019-09-25
Add automatic Jib support for Maven

chanseokoh:
Warning: I'm pretty new to Go.

If `pom.xml` exists and has Jib configured in it, the main automatically runs Jib. However, `Dockerfile` exists, it still takes the first precedence.

For Jib, it builds directly to GCR. It also overrides any configured credential helper to always use `docker-credential-gcloud`. Tested this only locally.


jamesward:
Thanks for this contribution and getting a conversation going around how to handle this.  Instead of running `mvn` as part of the button, I think it'd be preferable to have a `build` script like we were discussing in #85.  Because then we can easily support any project-based process that can create a container image.  And then we wouldn't need special handling for things like `mvnw` as well.  I have a PR for something pretty similar:
https://github.com/GoogleCloudPlatform/cloud-run-button/pull/86

But I'm awaiting review, so it could be totally wrong cause I don't really know what I'm doing.  :)

ahmetb:
I'm in favor of adding a few more "auto-build" modes –but it should not conflict with `pack` support. This one seems mostly harmless (I'll review later once I'm back from OOO).

If we can satisfy Java repos without having to write a build script, that would be most favorable.

ahmetb:
> Tested this only locally.

you can test it by building/pushing a **public** docker image + editing the Open in Cloud Shell URL we provide in readme with that image.

jamesward:
Right now the detection is based on running the Maven build and checking for the jib plugin.  Which I think it the root of the problem for pack fallback.  So for instance, we'd do this if there is a `pom.xml`:
1. Run the maven build to check for jib (and ideally also support mvnw which downloads maven on-the-fly instead of using the Cloud Shell provided one)
2. If jib plugin doesn't exist, run pack

The overhead of this is potentially significant (like a minute) which I why I favor an explicit build script instead of the detection attempt.

ahmetb:
Since I don’t know much Java: do we have to do maven build to “check for jib”? Isnt there more a static analysis we can do?

jamesward:
We can grep for it. :)  But there are probably a million corner cases.  I don't think it's too much hassle to have someone add their build command to an `app.json` and then we can keep that concern out of the button.  I don't think we want to head down the path of having all of the detections for all the ways a project can produce a container, in our code.

ahmetb:
Adding command to a script file is usually better than adding it to app.json. 
You almost never wanna deal with scripting semantics in a json/yaml file.

If we somehow said buildMode: jib explicitly in app.json, that would do fine.

jamesward:
We'd still have to figure out how to run the build, is it Maven or Gradle?  Does it use mvnw or gradlew?    And then we have to add another build mode for sbt's docker builder (like jib but different).  Repeat for all the ways that someone can create a container from their build.  It's a lot of work and a lot to maintain to avoid allowing the user to specify their build command.

Yeah, commands in JSON aren't great but they are pretty common practice and there is already precedent with app.json's postdeploy command support.  If we don't want to allow commands in app.json we could alternatively only support file references in the app.json scripts.

chanseokoh:
So I was also concerned about the slow Jib detection, which is due to Maven having to download a lot of artifacts into `~/.m2`. I think we can improve this by first doing a simple grep on pom.xml to see if `jib-maven-plugin` is configured, and if so, do the Maven detection. If we see `jib-maven-plugin`, there is a very good chance Jib is correctly configured in the default settings. The time spent on downloading the artifacts during detection won't be lost, because those artifacts would have to be downloaded anyway in the main Jib build later.

Personally, I hope the Cloud Run button does some special auto-handling for Jib projects so that repo owners can add the button easily without adding `app.json`. I also have thought about adding another URL parameter like `--disableJibDetection`. But having an extension to run arbitrary commands could be handy regardless.

chanseokoh:
Added a quick peek check to grep `<artifactId>jib-maven-plugin</artifactId>` in `pom.xml`. Projects without Jib won't get slowed down. Also nothing to lose for Jib projects, as I explained.

cmd/cloudshell_open/jib.go
```diff
@@ -0,0 +1,82 @@
...
+		if (stat.Mode() & 0111) != 0 {
+			if wrapper, err := filepath.Abs(filepath.Join(dir, "mvnw")); err == nil {
+				executable = wrapper
+			}
+		}
+	}
+
+	cmd := exec.Command(executable, args...)
+	cmd.Dir = dir
+	return *cmd
```
ahmetb: Try not to dereference a normally pointer type. Just return *Cmd from here

cmd/cloudshell_open/main.go
```diff
@@ -231,33 +232,49 @@ func run(c *cli.Context) error {
 		fmt.Println(infoPrefix + " FYI, running the following command:")
 		cmdColor.Printf("\tdocker build -t %s %s\n", parameter(image), parameter("."))
 	} else {
-		fmt.Println(infoPrefix + " Attempting to build this application with Cloud Native Buildpacks (buildpacks.io)...")
-		fmt.Println(infoPrefix + " FYI, running the following command:")
-		cmdColor.Printf("\tpack build %s --path %s --builder heroku/buildpacks\n", parameter(image), parameter(appDir))
+		jibMaven, err = jibMavenConfigured(appDir)
+		if err != nil {
+			return err
```
ahmetb: wrap with a message explaining it failed to detect jib

cmd/cloudshell_open/main.go
```diff
@@ -234,7 +234,7 @@ func run(c *cli.Context) error {
 	} else {
 		jibMaven, err = jibMavenConfigured(appDir)
 		if err != nil {
-			return err
+			return fmt.Errorf("filed to check if Jib is configured: %s", err)
```
ahmetb: failed*

cmd/cloudshell_open/jib.go
```diff
@@ -0,0 +1,82 @@
...
+			}
+		}
+	}
+
+	cmd := exec.Command(executable, args...)
+	cmd.Dir = dir
+	return cmd
+}
+
+func fileHasString(filePath string, pattern string) (bool, error) {
```
ahmetb: optional but this, seems unnecessary, you can just unfold into the call site.

cmd/cloudshell_open/jib.go
```diff
@@ -0,0 +1,82 @@
...
+	cmd := createMavenCommand(dir, "--batch-mode", "-Dmaven.test.skip=true",
+		"package", "jib:build", "-Dimage="+image, "-Djib.to.auth.credHelper=gcloud")
+	if b, err := cmd.CombinedOutput(); err != nil {
+		return fmt.Errorf("Jib Maven build failed: %v, output:\n%s", err, string(b))
+	}
+	return nil
+}
+
+func jibMavenConfigured(dir string) (bool, error) {
+	if _, err := os.Stat(filepath.Join(dir, "pom.xml")); err != nil {
```
ahmetb: if you actually unfold ioutil.ReadAll here, you can do os.IsNotExist() directly on its err, and therefore eliminate the os.Stat

jamesward:
I think before we merge this approach we do need to see how this is going to all play out with mvnw, gradle, and gradlew support.  Cause we need to see how the UX comes together.  We don't want to have a bunch of messages for things that don't pertain to most users.  e.g. I don't think we should tell every user that falls through to buildpacks that we couldn't detect a Jib project since I think that gets confusing in many cases.  And yeah, we'll need to do some testing since we don't have integration tests that validate the right decision flow (e.g. falling back to buildpacks).

chanseokoh:
> I don't think we should tell every user that falls through to buildpacks that we couldn't detect a Jib project 

I agree. If Jib is not detected, nothing about Jib will be logged–complete silence about Jib.
```
[ ? ] Choose a region to deploy this application: us-central1                                                       
[ ! ] Attempting to build this application with its Dockerfile...                                                                                                                                  
[ ! ] FYI, running the following command:                                                                           
        docker build -t gcr.io/chanseok-playground-new/cloud-run-hello .                                                          
[ ✓ ] Built container image gcr.io/chanseok-playground-new/cloud-run-hello                                                   
```

The repos I've been testing with are

1.  `--repo_url=https://github.com/GoogleCloudPlatform/cloud-run-hello.git`
2. `--repo_url=https://github.com/chanseokoh/jib-war-helloworld-test.git`
   - WAR project
3. `--repo_url=https://github.com/GoogleContainerTools/jib.git --dir=examples/dropwizard`
   - Official Jib repo. The example is under `examples/dropwizard`. The intended endpoint is at `...:8080/hello-world`.

> I think before we merge this approach we do need to see how this is going to all play out with mvnw, gradle, and gradlew support.

We can have more discussions, but what I've been thinking is that, we do the Jib Maven detection first and then Gradle. (Most repos will have either one anyway, unless they are primarily for examples.) Detecting and using a Maven wrapper is easy to support and already works in this PR. Gradle is no different in this regard. We've been supporting wrappers in Skaffold-Jib integration.

jamesward:
Cool. Didn't see the mvnw support.  That is good.  I'll give this a whirl on some test repos and see how it goes.

cmd/cloudshell_open/jib.go
chanseokoh: Cool. I've simplified the code around this.

chanseokoh:
> please also find a way to document this feature in README.md then we're good to go.

I'll try to think of something later after adding Jib Gradle support.

chanseokoh:
Move forward with this?

jamesward:
Sorry for the delay.  I'll get this reviewed asap.

jamesward:
So sorry for the delay.  Finally getting a chance to try this and so far it's been great.

I can't remember if it was supposed to work, but the Gradle repo I tried didn't work with jib:
https://github.com/jamesward/hello-play-gradle

chanseokoh:
This PR doesn't do anything about Gradle. Another PR later, when I get to that.

PR #95, opened 2019-10-02
Change memory to 256 from 512

geshan:
- [x] 256 can be a good start for most apps


geshan:
@googlebot I signed it!

jamesward:
Java apps are the exception. :(  I think we need to `app.json` this setting and make the default 256 after that.

geshan:
That's a good suggestion.

jamesward:
Closing this in favor of future `app.json` support for specifying memory.

jamesward:
BTW, the `app.json` configability is live: #175 And the default is back to 256.

PR #99, opened 2019-10-05
add GOOGLE_APPLICATION_CREDENTIALS env var

jamesward:


ahmetb:
Lgtm but this command is waaaay too long to be in 1 line. 

If you can take the time to break up with \ and indent, it would be great.

jamesward:
Let me know if that looks better.

ahmetb:
Looks good. Thanks.

PR #123, opened 2019-12-09
names must be lower case, numbers, or dashes - fixes #57

### LINKED ISSUE #57: Escape Docker Container Tag
The `name` from `app.json` can contain spaces and other chars which are not valid in the container tag.  So we need to escape them (which may already be in place for the Cloud Run service name).

Reproduce with:
https://console.cloud.google.com/cloudshell/editor?shellonly=true&cloudshell_image=gcr.io/cloudrun/button&cloudshell_git_repo=https://github.com/bkimminich/juice-shop.git

jamesward:


cmd/cloudshell_open/cloudrun.go
```diff
@@ -112,14 +113,24 @@ func tryFixServiceName(name string) string {
...
+
 	if len(name) > 63 {
 		name = name[:63]
 	}
+
 	for name[len(name)-1] == '-' {
 		name = name[:len(name)-1]
 	}
+
+	reg, _ := regexp.Compile("[^a-z0-9-]+")
```
ahmetb: use regexp.MustCompile so we crash the process in case regexp is incorrect.
ideally ignoring errors is rarely a good thing in Go.

ahmetb: you can also directly do MustCompile(..).ReplaceAllString(...)

jamesward: Thanks! As you can tell, I'm new to this Go thing :)

cmd/cloudshell_open/cloudrun.go
```diff
@@ -112,14 +113,24 @@ func tryFixServiceName(name string) string {
...
+
+	name = reg.ReplaceAllString(name, "-")
```
ahmetb: Please move this all the way up below ToLower.

For example if name is `šo`, this will result in `-o`, but if you convert this to `-o` early on, we will prefix it with `svc--o`, at least.

cmd/cloudshell_open/cloudrun_test.go
```diff
@@ -29,16 +29,18 @@ func Test_tryFixServiceName(t *testing.T) {
...
+			"a12345678901234567890123456789012345678901234567890123456789012"},
 		{"already at max length - no truncate",
 			"A12345678901234567890123456789012345678901234567890123456789012",
-			"A12345678901234567890123456789012345678901234567890123456789012",
+			"a12345678901234567890123456789012345678901234567890123456789012",
 		},
 		{"leading digit + trunc",
 			"0123456789012345678901234567890123456789012345678901234567890123456789",
 			"svc-01234567890123456789012345678901234567890123456789012345678"},
+		{"spaces", "foo bar", "foo-bar"},
```
ahmetb: can you add invalid chars to the beginning? also maybe generalize this as: `invalid chars` and use `"hello, world"` as the value. that actually tests `hello-world` and not `hello--world` (due to `+` in regexp)

PR #126, opened 2020-01-08
check that cloned dir does not add end up in PATH

jamesward:


cmd/cloudshell_open/main.go
```diff
@@ -139,6 +140,15 @@ func run(opts runOpts) error {
 		}
 	}
 
+	fullCloneDir, err := filepath.Abs(cloneDir)
```
ahmetb: let's handle "err".

cmd/cloudshell_open/main.go
```diff
@@ -139,6 +140,20 @@ func run(opts runOpts) error {
...
+
+	if err != nil {
+		return err
+	}
+
+	pathDirs := strings.Split(os.Getenv("PATH"), string(os.PathListSeparator))
+
+	for _, p := range pathDirs {
+		if strings.HasPrefix(p, fullCloneDir) {
```
ahmetb: this is not covering several cases (case-sensitivity etc). let's use this,

```
v, err := filepath.Rel(a,b)
if err != nil {
    // return err by wrapping with a message
} 
if !strings.HasPrefix(v, "..") {
   // fail
}
```

PR #139, opened 2020-01-30
Handle No Projects Correctly - fixes #138

### LINKED ISSUE #138: No Billing Error
We have handling for helping the user setup billing, but unfortunately it doesn't work in all scenarios.  In a new account I'm getting:
```
demo@cloudshell:~$ cloudshell_open --repo_url "https://github.com/jamesward-demo/hello-php.git" --page "editor" --git_branch "master"
[ ✓ ] Cloned git repository https://github.com/jamesward-demo/hello-php.git.
[ ✓ ] Queried list of your GCP projects
[ ? ] Would you like to use existing GCP project  to deploy this app? Yes
Error: error determining if billing account is linked: exit status 1. output=
ERROR: (gcloud.beta.billing.projects.describe) could not parse resource []
```

jamesward:
strings.Split on an empty string returns a slice of length 1 containing an empty string.  We need to detect this and return a new slice of length 0.

### DIFF
--- a/cmd/cloudshell_open/project.go
+++ b/cmd/cloudshell_open/project.go
@@ -43,8 +43,12 @@ func listProjects() ([]string, error) {
 	}
 	b = bytes.TrimSpace(b)
 	p := strings.Split(string(b), "\n")
-	sort.Strings(p)
-	return p, err
+	if len(p) == 1 && p[0] == "" {
+		return []string{}, err
+	} else {
+		sort.Strings(p)
+		return p, err
+	}
 }
 
 func promptProject(projects []string) (string, error) {
--- a/cmd/cloudshell_open/project.go
+++ b/cmd/cloudshell_open/project.go
@@ -41,14 +41,13 @@ func listProjects() ([]string, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to list projects: %+v, output:\n%s", err, string(b))
 	}
-	b = bytes.TrimSpace(b)
-	p := strings.Split(string(b), "\n")
-	if len(p) == 1 && p[0] == "" {
-		return nil, err
-	} else {
-		sort.Strings(p)
-		return p, err
+	s := string(bytes.TrimSpace(b))
+	if s == "" {
+		return nil, nil
 	}
+	p := strings.Split(s, "\n")
+	sort.Strings(p)
+	return p, err
 }
 
 func promptProject(projects []string) (string, error) {
--- a/cmd/cloudshell_open/project.go
+++ b/cmd/cloudshell_open/project.go
@@ -44,7 +44,7 @@ func listProjects() ([]string, error) {
 	b = bytes.TrimSpace(b)
 	p := strings.Split(string(b), "\n")
 	if len(p) == 1 && p[0] == "" {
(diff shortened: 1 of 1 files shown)


cmd/cloudshell_open/project.go
```diff
@@ -43,8 +43,12 @@ func listProjects() ([]string, error) {
 	}
 	b = bytes.TrimSpace(b)
 	p := strings.Split(string(b), "\n")
-	sort.Strings(p)
-	return p, err
+	if len(p) == 1 && p[0] == "" {
+		return []string{}, err
```
ahmetb: you can do `return nil, err`, nil is a valid value for slices.

cmd/cloudshell_open/project.go
```diff
@@ -43,8 +43,12 @@ func listProjects() ([]string, error) {
...
+	} else {
```
ahmetb: no need for `else{` because you return in previous block.

cmd/cloudshell_open/project.go
```diff
@@ -43,8 +43,12 @@ func listProjects() ([]string, error) {
 	}
 	b = bytes.TrimSpace(b)
 	p := strings.Split(string(b), "\n")
-	sort.Strings(p)
-	return p, err
+	if len(p) == 1 && p[0] == "" {
```
ahmetb: I think the cleaner course of action would be this:


```
b := string(bytes.TrimSpace(b))
if b == "" {
   return nil, nil
}

// split
// sort
// return
```

cmd/cloudshell_open/project.go
```diff
@@ -43,8 +43,12 @@ func listProjects() ([]string, error) {
...
+		return []string{}, err
+	} else {
+		sort.Strings(p)
+		return p, err
```
ahmetb: both `return x, err` should be `return x, nil`.
not sure why `err` is down here..

cmd/cloudshell_open/project.go
jamesward: Hmm...  Don't I need to return if the if didn't match?

ahmetb: I meant this, if you do

```
if cond {
   return
} else {
   logic
   return
}
```

several Go linting tools would tell you `else is superfluous` as it's equivalent to

```
if cond {
   return
}

logic
return
```

which is more readable.

ahmetb:
Consider having `fixes #xxx` in the commit/PR message (not title) so github can auto-close :)

cmd/cloudshell_open/project.go
jamesward: Ah. I see.  I'm used to expression-oriented ifs :)

jamesward:
> Consider having `fixes #xxx` in the commit/PR message (not title) so github can auto-close :)

It just needs to be in the first commit (which it was).  The PR picked the title from the first commit.

I've cleaned this up to your suggestion.

ahmetb:
Awesome, thanks! 
The CI will take ~forever. Feel free to merge yourself.

jamesward:
Thanks for your Go styling help :)

PR #146, opened 2020-03-17
Allow skipping builds and pre/post build hooks - fixes #85

### LINKED ISSUE #85: Build Script to Enable Jib and Other Builds

jamesward:


jamesward:
Tested with:
`--repo_url=https://github.com/jamesward/hello-uzhttp.git --git_branch=buttonify`

jamesward:
Oh, and whoops - for some reason my push went to `GoogleCloudPlatform` instead of my fork.  Will investigate why to try to avoid that in the future.

cmd/cloudshell_open/main.go
```diff
@@ -253,43 +253,39 @@ func run(opts runOpts) error {
...
-	exists, err := dockerFileExists(appDir)
-	jibMaven := false
-	if err != nil {
-		return err
-	}
-	if exists {
+	projectEnv := fmt.Sprintf("GOOGLE_CLOUD_PROJECT=%s", project)
+	regionEnv := fmt.Sprintf("GOOGLE_CLOUD_REGION=%s", region)
+	serviceEnv := fmt.Sprintf("K_SERVICE=%s", serviceName)
+	pathEnv := fmt.Sprintf("PATH=%s", os.Getenv("PATH"))
```
ahmetb: are the envs 100% the same as other hook? if so potentially worth extracting into a common method.

cmd/cloudshell_open/appfile.go
```diff
@@ -44,8 +44,9 @@ type hook struct {
 }
 
 type hooks struct {
-	PreCreate hook `json:"precreate"`
+	PreCreate  hook `json:"precreate"`
 	PostCreate hook `json:"postcreate"`
+	Build      hook `json:"build"`
```
ahmetb: I think this should be called PostBuild or PreBuild.
More verbose the better.

cmd/cloudshell_open/main.go
ahmetb: Ah ignore me, it seems like you just moved these above.

cmd/cloudshell_open/appfile.go
ahmetb: Wait, I think I got this wrong:

Does this override the build method (i.e. docker jib etc)?
If so:
- I think this is technically not a hook, but more like an "override". If - how does it know what image name to build as?

Please add README.md docs as well.

jamesward: Yeah, more details in https://github.com/GoogleCloudPlatform/cloud-run-button/issues/85#issuecomment-599804703

I added docs and two new env vars (image url and app dir).

ahmetb:
@jamesward the issues I see with this are:

- we are overusing _hooks_ section for _overrides_. maybe these two concepts will have different needs in the future. (we can split them by simply defining a `type buildConfig hook`)

- how does a "build" command know what image name/tag to build as?

- how does a "build" command know what to do? for example, do I just build an image onto the local docker engine and it's pushed for me ––or am I also responsible for pushing the image?

- could we potentially also do this as something like `skipBuild: true` and by adding a `preBuild` or `postBuild` hook which lets the user do it there?

jamesward:
> we are overusing hooks section for overrides. maybe these two concepts will have different needs in the future. (we can split them by simply defining a type buildConfig hook)

I'm fine with a new section.  How about `buildconfig` as a root JSON node?

> how does a "build" command know what image name/tag to build as?

There is a new env var `IMAGE_URL` which contains the expected container image url.  For tag, We've so far been tagless, but maybe we should include the git commit in the `IMAGE_URL`?

> how does a "build" command know what to do? for example, do I just build an image onto the local docker engine and it's pushed for me ––or am I also responsible for pushing the image?

Today the push still happens in our `cloudshell_open` but we could provide an `app.json` parameter to change that.  WDYT?

> could we potentially also do this as something like skipBuild: true and by adding a preBuild or postBuild hook which lets the user do it there?

That'd work too.  I don't have a strong preference either way.  WDYT?

ahmetb:
> That'd work too. I don't have a strong preference either way. WDYT?

That would help us maintain fewer things in general. People can probably even use precreate hook if `skipBuild: true` existed (although would seem hacky).

> There is a new env var `IMAGE_URL` which contains the expected container image url. For tag, We've so far been tagless, but maybe we should include the git commit in the `IMAGE_URL`?

IMO that's good enough. Tagless still works fine.
Should we be passing this env to all hooks or just this one? 

> I'm fine with a new section. How about `buildconfig` as a root JSON node?

We could probably just call it "build" as everything is technically config in that file. :)

jamesward:
Cool.  So how about:
```
{
    "build": {
        "skip": true
    },
    "hooks": {
        "prebuild": {
            "commands": [
                "./my-custom-build"
            ]
        },
        "postbuild": {
            "commands": [
                "./something"
            ]
        }
    }
}
```

> Should we be passing this env to all hooks or just this one?

I actually have a `postcreate` that needs `IMAGE_URL` (right now I just use convention) so I think we just send all the env vars to all hooks.

ahmetb:
Design LGTM.
I think we can use this `"build"` flag for bunch of overrides. For example, assume there's Dockerfile, but you want to use jib, we could implement build.jib=true in the future.

jamesward:
I've updated this to use the above schema.

README.md
```diff
@@ -62,7 +62,15 @@ For example:
...
     },
+    "build": {
+        "skip": false
+    },
     "hooks": {
+        "build": {
+            "commands": [
+                "./my-custom-build"
+            ]
+        },
```
ahmetb: this seems outdated now

README.md
```diff
@@ -91,9 +99,16 @@ Reference:
...
+  - `skip`: _(optional, default: `false`)_ skips the built-in build methods (`docker build`, `Maven Jib`, and `buildpacks`)
 - `hooks`: _(optional)_ Run commands in separate bash shells with the environment variables configured for the
- application and environment variables for `GOOGLE_CLOUD_PROJECT`, `GOOGLE_CLOUD_REGION`, and `K_SERVICE`. Command
- outputs are shown as they are executed.
+ application and environment variables `GOOGLE_CLOUD_PROJECT` (Google Cloud project), `GOOGLE_CLOUD_REGION`
+ (selected Google Cloud Region), `K_SERVICE` (Cloud Run service name), `IMAGE_URL` (container image URL), `APP_DIR`
+ (application directory). Command outputs are shown as they are executed.
+  - `prebuild`: _(optional)_ Runs the specified commands before running the built-in build methods
+    - `commands`: _(array of strings)_ The list of commands to run
+  - `postbuild`: _(optional)_ Runs the specified commands after running the built-in build methods
```
ahmetb: I think we can document here:

> You can use `IMAGE_URL` environment variable to build and push the required container image.

cmd/cloudshell_open/main.go
```diff
@@ -250,48 +250,54 @@ func run(opts runOpts) error {
...
-		fmt.Println(infoPrefix + " FYI, running the following command:")
-		cmdColor.Printf("\tdocker build -t %s %s\n", parameter(image), parameter("."))
-	} else {
-		jibMaven, err = jibMavenConfigured(appDir)
-		if err != nil {
-			return fmt.Errorf("failed to check if Jib is configured: %s", err)
-		}
-		if jibMaven {
+
+	if appFile.Build.Skip == nil || *appFile.Build.Skip == false {
```
ahmetb: lets rewrite this as to avoid extra indentation:

```
if (skip condition) {
    continue
} else if dockerfile {
   ...
} else if jib {
   ...
} ...
```

ahmetb:
Just did a quick pass. I'll probably do another but mostly LGTM.

jamesward:
Thanks.  Fixed all that up.  Let me know how it looks now.

README.md
```diff
@@ -23,7 +23,7 @@ Cloud](https://deploy.cloud.run/button.svg)](https://deploy.cloud.run/?git_repo=
...
     ```
 
-1. If the repo contains a `Dockerfile`, it will be built using the `docker build` command.  If the repo uses Maven for the build and it contains the [Jib plugin](https://github.com/GoogleContainerTools/jib/tree/master/jib-maven-plugin), then the container image will be built with Jib ([Jib Spring Boot Sample](https://github.com/GoogleContainerTools/jib/tree/master/examples/spring-boot)).  Otherwise, [CNCF Buildpacks](https://buildpacks.io/) (i.e. the `pack build` command) will attempt to build the repo ([buildpack samples][buildpack-samples]).
+1. If the repo contains a `Dockerfile`, it will be built using the `docker build` command.  If the repo uses Maven for the build and it contains the [Jib plugin](https://github.com/GoogleContainerTools/jib/tree/master/jib-maven-plugin), then the container image will be built with Jib ([Jib Spring Boot Sample](https://github.com/GoogleContainerTools/jib/tree/master/examples/spring-boot)).  Otherwise, [CNCF Buildpacks](https://buildpacks.io/) (i.e. the `pack build` command) will attempt to build the repo ([buildpack samples][buildpack-samples]).  Alternatively the build command can be defined in the `app.json` - see below.
```
ahmetb: > Alternatively the build command can be defined in the `app.json` - see below.

This looks outdated.

> Alternatively, you can skip build using the `build.skip` field (see below) and use a "prebuild" or "postbuild" hook to build and push the container image yourself.

Also we can paragraphize this entire bullet point like:

```
1. lorem ipsum

   lorem ipsum 

   lorem ipsum

1. another item
```

README.md
```diff
@@ -91,9 +104,18 @@ Reference:
   - `generator`, _(optional)_ use a generator for the value, currently only support `secret`
 - `options`: _(optional)_ Options when deploying the service
   - `allow-unauthenticated`: _(optional, default: `true`)_ allow unauthenticated requests
+- `build`: _(optional)_ Build configuration
+  - `skip`: _(optional, default: `false`)_ skips the built-in build methods (`docker build`, `Maven Jib`, and `buildpacks`)
```
ahmetb: let's append this:

> [...], given you should use a hook to build and push the image yourself.

README.md
```diff
@@ -91,9 +104,18 @@ Reference:
...
 - `hooks`: _(optional)_ Run commands in separate bash shells with the environment variables configured for the
- application and environment variables for `GOOGLE_CLOUD_PROJECT`, `GOOGLE_CLOUD_REGION`, and `K_SERVICE`. Command
- outputs are shown as they are executed.
+ application and environment variables `GOOGLE_CLOUD_PROJECT` (Google Cloud project), `GOOGLE_CLOUD_REGION`
+ (selected Google Cloud Region), `K_SERVICE` (Cloud Run service name), `IMAGE_URL` (container image URL), `APP_DIR`
+ (application directory). Command outputs are shown as they are executed.
+  - `prebuild`: _(optional)_ Runs the specified commands before running the built-in build methods. Use the `IMAGE_URL`
+  environment variable for the container storage coordinates.
+    - `commands`: _(array of strings)_ The list of commands to run
+  - `postbuild`: _(optional)_ Runs the specified commands after running the built-in build methods. Use the `IMAGE_URL`
```
ahmetb: > Use the `IMAGE_URL`
  environment variable for the container storage coordinates

the "coordinates" and "container storage" sound odd, never heard in this context.

What if we said:

> Use `IMAGE_URL` to determine the container image name you need to build and push.

cmd/cloudshell_open/main.go
```diff
@@ -250,48 +250,58 @@ func run(opts runOpts) error {
...
+
+	if appFile.Hooks.PreBuild.Commands != nil {
+		err = runScripts(appDir, appFile.Hooks.PreBuild.Commands, hookEnvs)
 	}
-	if exists {
+
+	skipBuild := appFile.Build.Skip != nil && *appFile.Build.Skip == true
+
+	if skipBuild {
+		// do nothing
```
ahmetb: do you think we should print an "info" or something? would this confuse end-users a lot?

PR #148, opened 2020-03-21
Start phasing out gcloud usages

ahmetb:
Start using client libraries for a few tasks (more to come).

Update dependencies while at it.

cc: @jamesward


ahmetb:
> Error processing tar file(exit status 1): write /google/devshell/editor/language_servers/java/server/plugins/org.apache.ant_1.10.7.v20190926-0324.jar: no space left on device

The disk space issue is hurting us badly, I suspect we can't escape this anymore. Maybe we should no longer build the docker image.

jamesward:
We could switch GitHub Actions for Cloud Build or some other CI.

go.mod
```diff
@@ -3,8 +3,14 @@ module github.com/GoogleCloudPlatform/cloud-run-button
 go 1.12
 
 require (
-	github.com/AlecAivazis/survey/v2 v2.0.1
-	github.com/briandowns/spinner v0.0.0-20190319032542-ac46072a5a91
-	github.com/fatih/color v1.7.0
-	google.golang.org/api v0.7.1-0.20190723010802-069bea57b1be
+	cloud.google.com/go v0.55.0 // indirect
```
jamesward: Just curious why these transitive dependencies need to be specified explicitly?

cmd/cloudshell_open/billing.go
```diff
@@ -15,28 +15,22 @@
...
+// GCP project ID.
 func checkBillingEnabled(projectID string) (bool, error) {
-	var o bytes.Buffer
-	var e bytes.Buffer
-	cmd := exec.Command("gcloud", "beta", "billing", "projects", "describe", "-q", "--format=json", projectID)
-	cmd.Stdout = &o
-	cmd.Stderr = &e
-	if err := cmd.Run(); err != nil {
-		return false, fmt.Errorf("error determining if billing account is linked: %+v. output=\n%s", err, e.String())
+	client, err := cloudbilling.NewService(context.TODO())
```
jamesward: What is `context.TODO()`?

go.mod
ahmetb: Thats what go mod commands currently do. 

They might change their mind later. 
¯\_(ツ)_/¯

cmd/cloudshell_open/billing.go
ahmetb: context is a rather complex topic in Go. It’s used for signaling cancellation in this case. 

Here we are not passing a particular context. So in the future we decide to propagate a context, we’d replace these.

ahmetb:
> We could switch GitHub Actions for Cloud Build or some other CI.

I like the failure emails from Github as well as public failure logs. As far as I know we give up on those when using GCB.

cmd/cloudshell_open/billing.go
jamesward: Ah cool.  Sounds kinda like an alternative to Java's ThreadLocal.  Thanks for the explanation.

jamesward:
> I like the failure emails from Github as well as public failure logs. As far as I know we give up on those when using GCB.

True. We could work together on a GCB for open source solution.  :)

ahmetb:
Superseded by #149.

PR #152, opened 2020-03-24
bump go version

jamesward:


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/GoogleCloudPlatform/cloud-run-button
 
-go 1.12
+go 1.14
```
ahmetb: we aren't supposed to update go.mod manually. This might be ok (given 1.12 -> 1.14 hasn't changed how go.mod works). Usually `go get -u ./...` is the way to have a refresh on this file for future reference.

ahmetb:
LGTM.

PR #166, opened 2020-04-29
Inherit os env vars for hook scripts

### LINKED ISSUE #159: Hook Commands Don't Have All Envs
I tried something that also needed `HOME`.  Should we enable all the host env vars to be passed to the hook commands?

ahmetb:
Fixes #159
cc: @jamesward

I hope it doesn't introduce new side-effects that previously were not there.


jamesward:
Thanks!

PR #167, opened 2020-04-30
use the context param to use a different buildpack builder image

jamesward:


ahmetb:
We can merge but let's move the option to the app.json in the future. :)

jamesward:
Agree :)  Here is my reminder: #169

PR #171, opened 2020-05-08
app manage url

jamesward:


ahmetb:
I think the original reason I chose that string was it gives more exposure to the platform and shows the user first the homepage which they can learn to navigate from (+it's a shorter string, fits into a line easier).

jamesward:
Yeah, that is true.  What about a line:
```
Manage all Cloud Run services:
https://blah
Manage this service:
https://asdf
```

Too much?

ahmetb:
I think the patch is fine as is, lets try this.

PR #172, opened 2020-05-14
increase timeout

jamesward:


ahmetb:
Thanks for fixing. What's the context here? Which API is taking longer?

jamesward:
Sorry for the rush job. Had to get this out super-quick.  So Java apps were not starting within the timeout on deploy, resulting in:
```
Error: the service did not become ready in 30s, check Cloud Console for logs to see why it failed
```
The revision did actually correctly deploy though.

The default timeout for startup with Cloud Run via `gcloud` is 4 minutes.  So I matched that.

PR #175, opened 2020-05-19
Add memory & cpu options

jamesward:


jamesward:
This leaves the default as 256Mi which is the same as `gcloud run deploy` and now that we have an easy way to increase it, I think it's best we stick with the upstream default.

cmd/cloudshell_open/deploy.go
```diff
@@ -133,13 +145,16 @@ func generateRevisionName(name string, objectGeneration int64) string {
...
-func patchService(svc *runapi.Service, envs map[string]string, image string) *runapi.Service {
+func patchService(svc *runapi.Service, envs map[string]string, image string, options options) *runapi.Service {
 	// merge env vars
 	svc.Spec.Template.Spec.Containers[0].Env = mergeEnvs(svc.Spec.Template.Spec.Containers[0].Env, envs)
 
 	// update container image
 	svc.Spec.Template.Spec.Containers[0].Image = image
 
+	// update the resources
+	svc.Spec.Template.Spec.Containers[0].Resources = optionsToResourceRequirements(options)
```
ahmetb, line 156: 🤔 should we be patching existing services with defaults?

imagine I deployed a previous revision with default (512mib), then I go  and update manually to 1GiB, then redeploy using button, now I get back 512Mib.

so I think we should omit this part of the change.

jamesward: Good point! Yeah, we should not.

cmd/cloudshell_open/appfile.go
```diff
@@ -36,7 +36,9 @@ type env struct {
 }
 
 type options struct {
-	AllowUnauthenticated *bool `json:"allow-unauthenticated"`
+	AllowUnauthenticated *bool  `json:"allow-unauthenticated"`
+	Memory               string `json:"memory"`
+	Cpu                  string `json:"cpu"`
```
ahmetb, line 41: Go convention for abbrs is to name this CPU

jamesward:
This is live. Test with: http://deploy.cloud.run/?git_repo=https://github.com/jamesward/cloud-run-button-jamesward.git&revision=bash_integration_tests&dir=integration_tests/memory-cpu

PR #187, opened 2020-07-10
correct source dir for docker build - fixes #186

### LINKED ISSUE #186: Docker Command Wrong
When doing a docker build, the command listed is:
```
docker build --tag gcr.io/blah/asdf .
```
But since the project is in a subdir, that command doesn't work if the user copy & pastes it.  Maybe it'd work to just change the listed command to:
```
docker build --tag gcr.io/blah/asdf asdf
```

jamesward:


cmd/cloudshell_open/main.go
```diff
@@ -306,7 +306,7 @@ func run(opts runOpts) error {
 	} else if !skipDocker && dockerFileExists {
 		fmt.Println(infoPrefix + " Attempting to build this application with its Dockerfile...")
 		fmt.Println(infoPrefix + " FYI, running the following command:")
-		cmdColor.Printf("\tdocker build -t %s %s\n", parameter(image), parameter("."))
+		cmdColor.Printf("\tdocker build -t %s %s\n", parameter(image), parameter(appDir))
```
ahmetb, line 309: I think this isn't quite what you meant in your issue.

`appDir = filepath.Join(cloneDir, opts.subDir)` will generate a verbose path but the goal of this parameter is to show the user something that works on their laptop as well as cloudshell.

Wouldn't it be better to do:

```
showDir = "."
if opts.subDir != "" { showDir = opts.subDir }
```

jamesward: The goal was really to have something the user could run in Cloud Shell for subsequent builds.  For this command:
```
(cd /tmp; ./cloudshell_open --repo_url=https://github.com/GoogleCloudPlatform/cloud-run-hello.git; rm -rf cloud-run-hello)
```
The output looks like:
```
docker build -t gcr.io/crb-test/cloud-run-hello cloud-run-hello
```

And in Cloud Shell when run from `~` the output should be the relative dir, and the command should be copy & pasteable for subsequent builds.

PR #189, opened 2020-07-15
Note that create hooks are skipped on existing services

glasnt:
Without this change, pre/post create hooks are silently skipped. 

The message as it is will not make sense if there aren't any create hooks are configured, but it's better than nothing.

### DIFF
--- a/cmd/cloudshell_open/appfile.go
+++ b/cmd/cloudshell_open/appfile.go
@@ -46,11 +46,11 @@ type hook struct {
 }
 
 type buildpacks struct {
-	Builder	string `json:"builder"`
+	Builder string `json:"builder"`
 }
 
 type build struct {
-	Skip *bool `json:"skip"`
+	Skip       *bool      `json:"skip"`
 	Buildpacks buildpacks `json:"buildpacks"`
 }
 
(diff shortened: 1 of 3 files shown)


jamesward:
I think this will be confusing to users who don't know what the hooks are.  Perhaps we only say something when there are hooks configured?  Or we better convey when a service is being updated vs created?

ahmetb:
If you need hooks for "update", I'm afraid the correct way to address this might be introduce new hooks (predeploy, postdeploy) that are called in both create/update. But we should probably update README.md to say create-hooks aren't going to be invoked on redeploys.

jamesward:
The `prebuild` and `postbuild` hooks can be used for cases where hooks are needed on create & update.

jamesward:
Oh, but maybe we don't have a way to do a `postupdate` hook.  I haven't needed that myself, but maybe there is a use case?

glasnt:
To solve my use case (remind me that I'm updating rather than creating), could I submit a different PR that at least adds an information note when [existingService is not nil](https://github.com/GoogleCloudPlatform/cloud-run-button/blob/master/cmd/cloudshell_open/main.go#L258)?

jamesward:
Yeah, I think we either do something easy, like:
`The service blah exists and will be updated` (before we do the `gcloud run deploy`)
Or we get fancy and say:
`The service blah exists, update it? [Y|n]`
Then if they say `n` ask them for a new service name.

PR #191, opened 2020-07-22
fix build indicator

### LINKED ISSUE #182: No Busy Spinner on Build
I think the busy indicator used to spin during the build but it doesn't anymore.  Might be something I broke.

ahmetb:
fixes #182
cc: @jamesward


jamesward:
Thank you!

Issue #112, opened 2019-11-25
Both Cloud Run Button and Heroku Button - app.json Clash


SudharakaP:
I am trying to add the Cloud Run Button to [one of our repositories](https://github.com/jhipster/jhipster-registry) which has the [Heroku Deploy Button](https://devcenter.heroku.com/articles/heroku-button) which also uses an [`app.json` file](https://devcenter.heroku.com/articles/heroku-button). 

This creates problems when trying to add the Cloud Run button to the same repository as the Cloud Run also assumes an `app.json` file and the Heroku `app.json` file which is different in format will crash the Cloud Run when trying to deploy. 

Is there a possibility to configure the name of the `app.json` file for example so that it doesn't clash with other cloud providers such as Heroku? :smile:

ahmetb:
If you can mention the error, we can help it not clash (by perhaps introducing a placeholder for fields that aren't supported).

jamesward:
Yeah, our `app.json` parsing is strict so likely we just need more ignore flags.  I think this is the file that is failing:
https://github.com/SudharakaP/jhipster-registry/blob/add-gcp-cloud-run-deploy-button/app.json

I can investigate.

SudharakaP:
@ahmetb : Thanks for the quick reply. The first error is; 

`Error: error attempting to read the app.json from the cloned repository: failed to parse app.json file: failed to parse app.json: json: cannot unmarshal string into Go struct field env.required of type bool`

Now this is caused by the [following line of `app.yaml`](https://github.com/SudharakaP/jhipster-registry/blob/dbe7687806d1cb1f49f2211d11d57938c4ed72d9/app.json#L8). When you remove the quotes around `true` it passes through that but gives the following error; 

`Error: error attempting to read the app.json from the cloned repository: failed to parse app.json file: failed to parse app.json: json: unknown field "buildpacks"`

due to the [buildpacks argument in the `app.yaml` file](https://github.com/SudharakaP/jhipster-registry/blob/dbe7687806d1cb1f49f2211d11d57938c4ed72d9/app.json#L23)

Hope this helps. Let me know if you need any further details. 😄

ahmetb:
We can try and fix `buildpacks` argument, but I assume this is not a good idea. Because that value is critical to how the app is built. And we better not ignore it.

[...]

PR #193, opened 2020-07-27
add formation ignore app.json field for #112

jamesward:


cmd/cloudshell_open/appfile.go
```diff
@@ -71,12 +76,13 @@ type appFile struct {
...
-	IgnoredRepository  string   `json:"repository"`
-	IgnoredWebsite     string   `json:"website"`
-	IgnoredStack       string   `json:"stack"`
+	IgnoredDescription string               `json:"description"`
+	IgnoredKeywords    []string             `json:"keywords"`
+	IgnoredLogo        string               `json:"logo"`
+	IgnoredRepository  string               `json:"repository"`
+	IgnoredWebsite     string               `json:"website"`
+	IgnoredStack       string               `json:"stack"`
+	IgnoredFormation   map[string]formation `json:"formation"`
```
ahmetb, line 85: consider just saying `interface{}` here (means any type) so you don't have to define `type formation`

jamesward: oh! cool. didn't know about that.

jamesward:
This is better now. thanks for the tip!

ahmetb (approved):

PR #195, opened 2020-07-27
add region and project params for easier testing

jamesward:


cmd/cloudshell_open/main.go
```diff
@@ -68,6 +70,9 @@ func init() {
 	flags.StringVar(&opts.gitBranch, flGitBranch, "", "(optional) branch/revision to use from the git repo")
 	flags.StringVar(&opts.subDir, flSubDir, "", "(optional) sub-directory to deploy in the repo")
 	flags.StringVar(&opts.context, flContext, "", "(optional) arbitrary context")
+	flags.StringVar(&opts.project, flProject, "", "(optional) the GCP project")
```
ahmetb, line 73: Pls note these are: (testing-only). 

I've been refraining from creating cli opts for things that don't come from cloudshell.

I'd probably consider hidden/undocumented env vars for this purpose but up to you. Since we don't offer a stable API, we can remove these later if we don't like.

jamesward: Cool.  For `project` we could reuse the `GOOGLE_CLOUD_PROJECT` env var.  For `region` probably `GOOGLE_CLOUD_REGION` to match the hooks.  WDYT?

ahmetb: yeah if you made up your mind about env vars, we could totally do those.

jamesward: Cool. Switched to env vars.

PR #199, opened 2020-09-15
do not show redirect form if repo is specified - fixes #197

### LINKED ISSUE #197: Redirector With Explicit git_url From GitHub Loses Param
I have a repo that hard codes the `git_repo` and `revision` params in the button:
https://github.com/jamesward/hello-java/tree/jdk15

But the redirector gets confused because it sees the referer as `https://github.com` and so it asks the user to still specify the repo.

jamesward:


cmd/redirector/main.go
```diff
@@ -67,7 +67,7 @@ func redirect(w http.ResponseWriter, req *http.Request) {
 	referer := req.Header.Get(hdrReferer)
 
 	// TODO(ahmetb): remove once https://github.community/t/chrome-85-breaks-referer/130039 is fixed
-	if referer == "https://github.com/" {
+	if referer == "https://github.com/" && repoParam != "" {
```
ahmetb, line 70: Shouldn't it be repoParam=="" ?

If there's the repo parameter specified, we don't want to show the form?

jamesward: Eeek. Yeah, nice catch.  I had tested this but failed to test it correctly.

ahmetb (approved):
I forgot we had tests for that. I would have added more tests at the time of the workaround otherwise.

PR #203, opened 2020-12-10
add --use-http2 flag support - fixes #202

### LINKED ISSUE #202: Add --use-http2 Option
```
"options": {
  "use-http2": true
}
```

If `true` set `--use-http2`. If `false` then set `--no-use-http2`. If not set then omit.

jamesward:


jamesward:
Note that today the project must have the http2 alpha enabled.

README.md
```diff
@@ -69,7 +69,8 @@ For example, a fully populated `app.json` file looks like this:
         "allow-unauthenticated": false,
         "memory": "512Mi",
         "cpu": "1",
-        "port": "80"
+        "port": "80",
+        "http2": true
```
ahmetb, line 73: let's set example to `"Http2":false`

cmd/cloudshell_open/appfile.go
```diff
@@ -40,6 +40,7 @@ type options struct {
 	Memory               string `json:"memory"`
 	CPU                  string `json:"cpu"`
 	Port                 int    `json:"port"`
+	Http2                *bool  `json:"http2"`
```
ahmetb, line 43: Go vet/lint might fail if we don't call this `HTTP2` uppercase.

cmd/cloudshell_open/deploy.go
```diff
@@ -82,6 +82,20 @@ func optionsToResourceRequirements(options options) *runapi.ResourceRequirements
 	return &runapi.ResourceRequirements{Limits: limits}
 }
 
+func optionsToContainerSpec(options options) *runapi.ContainerPort {
+	var containerPortName = "http1"
+	if options.Http2 != nil && *options.Http2 {
+		containerPortName = "h2c"
+	}
+
+	var containerPort = 8080
```
ahmetb, line 91: I'll check with the team if this is allowed:

```
ports:
- name: h2c
```

which doesn't specify PORT and lets system assign 8080.

in Kubernetes this is not allowed, but on Cloud Run it might be.

otherwise we'd be hardcoding 8080, which actually isn't part of our runtime/container contract. so if Cloud Run changes 8080 default to something else tomorrow we'll be diverging from that.

cmd/cloudshell_open/main.go
```diff
@@ -440,6 +440,16 @@ func optionsToFlags(options options) []string {
...
+	if options.Http2 != nil && *options.Http2 == false {
+		http2Setting := "--no-use-http2"
+		flags = append(flags, http2Setting)
+	}
+
+	if options.Http2 != nil && *options.Http2 == true {
+		http2Setting := "--use-http2"
+		flags = append(flags, http2Setting)
+	}
+
```
ahmetb, line 452: rewrite as:

```
if options.Http2 != nil {
  if *options.Http2 { ... }
  else { ... }
}
```

no need to define `http2Setting` also.

ahmetb (changes requested):
thanks for getting to this,
it was on my list to do soon.

README.md
jamesward:
```suggestion
        "http2": false
```

cmd/cloudshell_open/deploy.go
jamesward: From what I could tell looking at Cloud Run YAML is that `8080` was the default value for all services. But yeah, I'm not sure how that translates to using `PORT`.

ahmetb:
One of the tests is failing:
```
=== RUN   TestGitCheckout
Error:     clone_test.go:110: git checkout failed: exit status 1, output:
        error: pathspec 'main' did not match any file(s) known to git
        
--- FAIL: TestGitCheckout (0.01s)
```
we can go merge but I'll try to investigate why.

ahmetb:
Also somehow go1.15 fails to compile at 0e5f62b:

```
# github.com/GoogleCloudPlatform/cloud-run-button/cmd/cloudshell_open
cmd/cloudshell_open/deploy.go:158:10: conversion from int to string yields a string of one rune, not a string of digits (did you mean fmt.Sprint(x)?)
```

jamesward:
I believe the test is failing in Actions due to an older version of `git` where the default branch isn't `main`.  Locally the test passes. Not sure how to workaround that.

Where are we using `go1.15` ?

ahmetb:
> Where are we using `go1.15` ?

on my development machine :) I pushed a commit fixing that.

we can t.Skip() the clone test maybe for the time being.

ahmetb:
I believe we'll need to add

```
git config --global init.defaultBranch main
```

to the workflow configuration. I'll try that now.

jamesward:
I added a param to `git init` which should explicitly set the branch to `main` (assuming the git version is high enough to have that flag)

ahmetb:
I think the config outside the code is easier to cleanup later. I fixed that in 6a416b6 outside the code. (you now force-pushed over it/)

Also I think that particular test statement checking out to default branch isn't that useful. Maybe we can remove that, too.

jamesward:
Whoops. Fun working on the same branch trying to fix the same thing.  I'll let you take over :)

ahmetb:
Yeah it seems both options were fine.
Tests seem to be passing. We can merge.

jamesward:
Cool. Thanks! Go ahead and merge it

ahmetb:
I hope this gets picked up soon, it seems like we could use this in the announcement blog post.

jamesward:
I can ask for a deploy

jamesward:
This change is now deployed.

PR #209, opened 2021-01-14
provide env ordering via "order" key

ahmetb:
Introducing an optional "order" key to the env object.

The sorting takes a look at if "order" field is specified or not. If not,
the key is prompted last. If specified, it sorts them by the value of the
"order" field.

The order is still non-deterministic if env objects don't specify an
"order" value. (i.e. this patch still doesn't preserve map order).

Fixes #205.
cc: @jamesward

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,10 @@ For example, a fully populated `app.json` file looks like this:
         },
         "APP_SECRET": {
             "generator": "secret"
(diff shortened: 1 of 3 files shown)


jamesward:
Awesome! Thank you.

jamesward (approved):

PR #211, opened 2021-01-20
implement signaling clone status

### LINKED ISSUE #210: The Cloud Shell Editor Does Not Open
Because we override `cloudshell_open` a message is never sent to Cloud Shell to tell it the git repo has been cloned.  This causes the editor to hang if a user tries to open it.
image

To fix this we need to send a tcp message that tells Cloud Shell the git repo has been cloned (in bash form):
```
echo -e "42\n[null,null,null,[null,null,null,null,[1]]]\n" > /dev/tcp/localhost/8998
```
Replace the `1` above with `0` if the clone failed.  `42` is the length.

ahmetb:
fixes #210
bug:178009327
cc: @jamesward


jamesward (approved):

jamesward:
Thanks for the quick turnaround!

PR #214, opened 2021-01-26
if http2 option is true then use gcloud alpha - fixes #213

### LINKED ISSUE #213: Add alpha to deploy command when http2 option is true
Is there a way to execute the beta version of the gcloud application when clicking the cloud run button?

jamesward:


cmd/cloudshell_open/main.go
```diff
@@ -374,11 +374,16 @@ func run(opts runOpts) error {
 		}
 	}
 
+	deployCommand := "\tgcloud run deploy %s"
+	if appFile.Options.HTTP2 != nil && *appFile.Options.HTTP2 == true {
+		deployCommand = "\tgcloud alpha run deploy %s"
```
ahmetb, line 379: should be `beta`.

jamesward: Maybe Cloud Shell doesn't have the latest cause I get:
```
ERROR: (gcloud.beta.run.deploy) The `--use-http2` flag is only supported in the alpha release track on the fully m
anaged version of Cloud Run. Use `gcloud alpha` to set `--use-http2` on Cloud Run (fully managed). Alternatively, 
specify `--platform gke` or run `gcloud config set run/platform gke`.
```

ahmetb: Sounds like a UX issue between Cloud Shell <=> Cloud SDK. :( 
I'd say let's do `beta` anyway as the issue will fix itself over time.

ahmetb: Might be also wise to add:

    // TODO remove when --use-http2 graduates to GA

jamesward: Yeah, I wonder when Cloud Shell pulls the latest gcloud bits?  I'll find that out since it'd be ideal if they do that when they pull our new image.

jamesward: Cloud Shell's image has been updated to the latest gcloud. So next build we do should get that change.

PR #219, opened 2021-04-14
experiment: instrumentless for users without projects or billing accounts

jamesward:
Starting as a draft since we need to setup the actual instrumentless event.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -63,10 +63,9 @@
 
 ## Test Instrumentless
 
+Test getting a coupon from the instrumentless API:
 ```
-go build -o /tmp/instrumentless_test ./cmd/instrumentless_test
-
-/tmp/instrumentless_test YOUR_EVENT $(gcloud auth print-access-token)
+go run ./cmd/instrumentless_test YOUR_EVENT $(gcloud auth print-access-token)
 ```
 
 ## Contributor License Agreement
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,6 +61,14 @@
       --repo_url=https://github.com/GoogleCloudPlatform/cloud-run-hello.git"
     ```
 
+## Test Instrumentless
+
+```
+go build -o /tmp/instrumentless_test ./cmd/instrumentless_test
+
+/tmp/instrumentless_test YOUR_EVENT $(gcloud auth print-access-token)
+```
+
 ## Contributor License Agreement
 
 Contributions to this project must be accompanied by a Contributor License
--- a/cmd/cloudshell_open/main.go
+++ b/cmd/cloudshell_open/main.go
@@ -43,7 +43,7 @@ const (
 	reauthCredentialsWaitTimeout     = time.Minute * 2
 	reauthCredentialsPollingInterval = time.Second
 
-	instrumentlessEvent = "cloudrunbutton"
+	instrumentlessEvent = "crbutton"
 )
 
 var (
--- a/cmd/cloudshell_open/main.go
+++ b/cmd/cloudshell_open/main.go
@@ -20,6 +20,8 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"github.com/GoogleCloudPlatform/cloud-run-button/cmd/instrumentless"
+	"google.golang.org/api/transport"
 	"os"
 	"path/filepath"
 	"strings"
@@ -41,7 +43,7 @@ const (
 	reauthCredentialsWaitTimeout     = time.Minute * 2
 	reauthCredentialsPollingInterval = time.Second
 
-	projectCreateURL = "https://console.cloud.google.com/cloud-resource-manager"
+	instrumentlessEvent = "cloudrunbutton"
 )
 
 var (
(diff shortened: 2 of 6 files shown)


cmd/cloudshell_open/main.go
```diff
@@ -41,7 +43,7 @@ const (
 	reauthCredentialsWaitTimeout     = time.Minute * 2
 	reauthCredentialsPollingInterval = time.Second
 
-	projectCreateURL = "https://console.cloud.google.com/cloud-resource-manager"
+	instrumentlessEvent = "crbutton"
```
ahmetb, line 46: if these aren’t pre-registered, can we do cloud-run-button?

cmd/cloudshell_open/main.go
```diff
@@ -228,14 +235,24 @@ func run(opts runOpts) error {
...
+		coupon, err := instrumentlessCoupon()
+		if err != nil {
+			return err
+		}
+
 		fmt.Print(errorPrefix+" "+
-			warningLabel.Sprint("GCP project you chose does not have an active billing account!")+
-			"\n  1. Visit "+linkLabel.Sprint(projectCreateURL),
-			"\n  2. Associate a billing account for project "+parameterLabel.Sprint(p),
+			warningLabel.Sprintf("Project '%s' does not have an active billing account!", project)+
```
ahmetb, line 244: I think changing this text is bad for existing GCP users with an active billing account?

cmd/cloudshell_open/main.go
```diff
@@ -549,3 +566,19 @@ func isSubPath(a, b string) (bool, error) {
 	}
 	return !strings.HasPrefix(v, ".."+string(os.PathSeparator)), nil
 }
+
+func instrumentlessCoupon() (*instrumentless.Coupon, error) {
+	ctx := context.TODO()
+
+	creds, err := transport.Creds(ctx)
+	if err != nil {
+		return nil, err
```
ahmetb, line 575: do you mind wrapping the errors? they help with debugging  when users submit an error msg.

cmd/instrumentless/instrumentless.go
```diff
@@ -0,0 +1,65 @@
...
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/ioutil"
+
+	"net/http"
+)
+
+type Coupon struct {
+	Url string `json:"url"`
```
ahmetb, line 27: set to URL otherwise golint would complain

cmd/instrumentless/instrumentless.go
```diff
@@ -0,0 +1,65 @@
...
+	body, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	coupon := Coupon{}
+	err = json.Unmarshal(body, &coupon)
+	if err != nil {
+		return nil, err
+	}
```
ahmetb, line 62: you can do:

json.NewDecoder(res.Body).Decode(&coupon) and eliminate some lines

cmd/instrumentless_test/main.go
```diff
@@ -0,0 +1,42 @@
+// Copyright 2019 Google LLC
```
ahmetb, line 1: can't this be a small bash script or a curl command we provide in maintainer docs?

ahmetb (approved):

CONTRIBUTING.md
```diff
@@ -61,6 +61,14 @@
       --repo_url=https://github.com/GoogleCloudPlatform/cloud-run-hello.git"
     ```
 
+## Test Instrumentless
+
+```
+go build -o /tmp/instrumentless_test ./cmd/instrumentless_test
+
+/tmp/instrumentless_test YOUR_EVENT $(gcloud auth print-access-token)
```
ahmetb, line 69: you can do:

    go run ./dir arg1 arg2

cmd/cloudshell_open/main.go
jamesward: This one is already registered as `crbutton`.

cmd/cloudshell_open/main.go
jamesward: You thinking we should make it clear they could use an existing billing account?  Like:

If they have one existing billing account:
```
Project 'foo' does not have an active billing account!
Would you like to use the existing `asdf` billing account with project `foo`? [Y/n]
n
Create a new billing account from a starter coupon:
https://console.cloud.google.com/usecoupon?asdf

Once you've completed creating the new billing account and associating it with your project, press [enter] to continue.
```

If they don't have an existing billing account, omit the option to use the existing one.

If they have more than one billing account, then:
```
Would you like to use an existing billing account with project `foo`? [Y/n]
Y
Select the billing account to use:
...
```

I'll see about doing the linking via the API.  Cause that'd make this nicer.

cmd/instrumentless_test/main.go
jamesward: It is nice to have some way to test `GetCoupon` manually. Not sure what the better way is though.

cmd/cloudshell_open/main.go
ahmetb: Yeah that's what I was proposing.
Most users with billing accounts don't need instrumentless at all.

I think we send users to the Console page for linking because I don't think there is an API for it.

cmd/instrumentless/instrumentless.go
```diff
@@ -50,15 +48,10 @@ func GetCoupon(event string, bearerToken string) (*Coupon, error) {
...
-	if err != nil {
-		return nil, err
-	}
-
 	coupon := Coupon{}
-	err = json.Unmarshal(body, &coupon)
+	err = json.NewDecoder(res.Body).Decode(&coupon)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("could not get a coupon: %v", err)
```
ahmetb, line 54: I think this error would be:

> failed to read and parse the response: %v

and "could not get a coupon" would be at the call site of this fn

ahmetb (approved):
Looks good to me.

PR #220, opened 2021-05-17
handling for invalid trycgp url and option to create a project

jamesward:


cmd/cloudshell_open/main.go
```diff
@@ -611,12 +598,24 @@ func instrumentlessCoupon() (*instrumentless.Coupon, error) {
...
 	coupon, err := instrumentlessCoupon()
-	if err != nil {
-		return fmt.Errorf("could not get instrumentless coupon: %v", err)
+
+	if err != nil || coupon == nil {
+		fmt.Println(infoPrefix + " Create a new billing account:")
+		fmt.Println("  " + linkLabel.Sprint(billingCreateURL))
+		fmt.Println(questionPrefix + " " + "Once you're done, press " + parameterLabel.Sprint("Enter") + " to continue: ")
+
+		if _, err := bufio.NewReader(os.Stdin).ReadBytes('\n'); err != nil {
```
ahmetb, line 607: isnt it kinda weird this is inside promptInstrumentless?
wouldn’t it be better if we do this if promptInstrumentless returns err?

ahmetb: i.e. I think it was fine where it was?

cmd/cloudshell_open/main.go
```diff
@@ -611,12 +598,24 @@ func instrumentlessCoupon() (*instrumentless.Coupon, error) {
...
+			return err
+		}
+
+		return nil
 	}
 
-	fmt.Println(infoPrefix + " Apply a starter coupon to create a billing account:" +
-		"\n  " + linkLabel.Sprint(coupon.URL))
+	code := strings.TrimPrefix(coupon.URL, trygcpURL+";code=")
```
ahmetb, line 614: you can probably do URL.Parse then get .Path and trim that if you want (in case base domain changes in the future etc) but this is fine for now.

cmd/cloudshell_open/main.go
jamesward: I realized that it needs to be in the two places we call `promptInstrumentless` but I could create another method that wraps this one.  wdyt?

ahmetb: honestly anything works, just trying to clean up the code bit. we don't unit-test these but nesting too much logic here is bad for future readability potentially.

Issue #108, opened 2019-11-15
[Feature Request] Add configuration for Cloud Run CLI flags


averikitsch:
Setting flags such as max instance and memory size from the configuration json.

ahmetb:
Related: was discussed a bit here https://github.com/GoogleCloudPlatform/cloud-run-button/pull/89#issuecomment-541079900

I think overall we need some property bag objects in app.json to control various aspects. Ideally, not a flat list, as that can result in gigantic objects over time as we add more flags to gcloud. So maybe some categorization of such objects can help.

jamesward:
Mem & cpu options are now available via `app.json` See: #175

benwinding:
Would also be great to have the `concurrency` flag too if possible!

glasnt:
Many flags were added in #228; there needs to be some thought as to how to make these flags easier to add in the future; maybe directly passing through to gcloud run deploy?

PR #228, opened 2022-02-28
Add max-instances and concurrency options

### LINKED ISSUE #227: support for settings instances to 1
how to set the number of instances to only 1 for apps that don't work when scaled, eg: https://github.com/matti/http-doom/

https://github.com/matti/http-doom/issues/6

glasnt:
Only adds options if defined by user (otherwise doesn't pass a default value, allowing downstream systems to specify)

Resolves #227

Addresses request in #108, but doesn't scale the solution as suggested.

Tested in isolation, pending automated testing

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,9 @@ For example, a fully populated `app.json` file looks like this:
         "memory": "512Mi",
         "cpu": "1",
         "port": 80,
-        "http2": false
+        "http2": false,
+        "concurrency": 80,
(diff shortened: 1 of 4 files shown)


jamesward (approved):

jamesward:
Thanks! This looks good to me.  Unfortunately we don't have any automated integration tests.  So I'll let you merge this after when you are ready / feel good with sufficient manual testing.

PR #246, opened 2023-12-01
use paketobuildpacks for custom builder validation

glasnt:
Cloud Run button supports supplying a custom builder, rather than using the default "gcr.io/buildpacks/builder". 

The buildpacks-builder test was failing due to [Heroku buildpacks](https://hub.docker.com/r/heroku/buildpacks) no longer having a tagged latest (presumably this worked before), and have also have been deprecated, replaced with [heroku/builder](https://hub.docker.com/r/heroku/builder/tags). However, both of these versions are not compatible with the version of pack being used (e.g.`ERROR: failed to build: Builder 'heroku/builder:22' is incompatible with this version of pack`)

So instead, replace the use of heroku with another buildpack. From the `pack builder suggest` output, `paketobuildpacks/builder` is an option. Using the `:full` image since the sample application used is PHP, and the `:tiny` version doesn't support this.

### DIFF
--- a/tests/buildpacks-builder/app.json
+++ b/tests/buildpacks-builder/app.json
@@ -1,7 +1,7 @@
 {
   "build": {
     "buildpacks": {
-      "builder": "heroku/buildpacks"
+      "builder": "paketobuildpacks/builder:full"
     }
   }
 }


jamesward:
Thanks @glasnt! Can you see why the integration test is failing?  I don't have access to the GCP project for it.

jamesward:
I think the integration test is failing because it is fetching the `master` branch for the tests instead of the branch the PR is on:
```
Command './cloudshell_open --repo_url=https://github.com/GoogleCloudPlatform/cloud-run-button --git_branch=master --dir=tests/buildpacks-builder' returned 1
```

The actual error is:
```
ERROR: failed to build: failed to fetch builder image 'index.docker.io/heroku/buildpacks:latest': image 'index.docker.io/heroku/buildpacks:latest' does not exist on the daemon: not found
```

So I think this PR fixes things but just isn't being fetched to test the fix.  If that sounds right we can just merge this.

glasnt:
Correct: the integration test only works for non-testing related changes. I've setup a separate Cloud Build trigger to allow customisation of the branch where tests are sourced from (integration-tests-MANUAL). 

You should have access to those test results to verify the change: https://github.com/GoogleCloudPlatform/cloud-run-button/pull/246/checks?check_run_id=19266836330

jamesward (approved):

jamesward:
Awesome.  Thanks @glasnt!

PR #273, opened 2024-08-22
feat: implement artifact registry integration

glasnt:
This replaces the implicit Container Registry repository logic with custom Artifact Registry logic. 

Namely, it either uses the existing `cloud-run-source-deploy` (a common name for the automatically created repository in `gcloud`) in the selected region, or create it for you. 

Presumes (as with all current operations) that this code will be invoked as a user within a Cloud Shell, which will by default be an owner of the selected project, so no additional permissions need to be granted. 

Also updates the base Go version and GitHub action step versions


cmd/cloudshell_open/artifactregistry.go
```diff
@@ -0,0 +1,63 @@
...
+	//Check for existing repo
+	client, err := artifactregistry.NewClient(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to create artifact registry client: %w", err)
+	}
+
+	req := &artifactregistrypb.GetRepositoryRequest{
+		Name: repoFull,
+	}
+	var existingRepo, _ = client.GetRepository(ctx, req)
```
subfuzion, line 42: Nit: a short variable declaration here would be idiomatic and more consistent with the other declarations in this function.

subfuzion (dismissed):
Just one nit, but otherwise LGTM

.github/workflows/ci.yml
```diff
@@ -7,11 +7,13 @@ jobs:
     - name: Free disk space
       run: |
         docker system prune -a -f
+        sudo rm -rf /usr/local/lib/android
```
grayside, line 10: question: Would this work without sudo?

cmd/cloudshell_open/artifactregistry.go
grayside, line 42: question: If the repo happens to exist but there was an error in checking if it does, will the repo creation below fail? Is there a risk of overwriting settings?

cmd/cloudshell_open/cloudrun.go
```diff
@@ -115,9 +115,9 @@ func envVars(project, name, region string) (map[string]struct{}, error) {
...
 // as:
 //
-//   * names with a leading non-letter (e.g. digit or '-') are prefixed
-//   * names over 63 characters are truncated
-//   * names ending with a '-' have the suffix trimmed
+//   - names with a leading non-letter (e.g. digit or '-') are prefixed
+//   - names over 63 characters are truncated
+//   - names ending with a '-' have the suffix trimmed
 func tryFixServiceName(name string) string {
 	if name == "" {
```
subfuzion, line 122: Being very nitpicky here, but this seems like this would be an erroneous situation, and a function name that starts with `try` sort of implies to me that it should return errors for unexpected things (maybe just baggage from working with too many languages with try/catch), but otherwise, why even guard here? Either we never expect this to happen or else it's probably better to fail here than when creating the image fails with an empty service name.

Maybe better if we have this?

```go
func tryFixServiceName(name string) (string, error) {
        if name == "" {
                return "", fmt.Errorf("service name can't be empty")
        }
```

cmd/cloudshell_open/main.go
```diff
@@ -291,14 +292,24 @@ func run(opts runOpts) error {
...
+	if err != nil {
+		return err
+	}
+
 	repoName := filepath.Base(appDir)
 	serviceName := repoName
 	if appFile.Name != "" {
 		serviceName = appFile.Name
 	}
 	serviceName = tryFixServiceName(serviceName)
```
subfuzion, line 310: Based on previous suggestion about the function returning an error, then this should check for err != nil...

subfuzion (commented):
One more suggestion (but feel free to ignore it).

.github/workflows/ci.yml
subfuzion: I saw this and got tripped up by the docker command before it (github docker containers run as root), but then realized this is running on the vm itself, which would require sudo (for removing a directory under `/usr`): https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#administrative-privileges

cmd/cloudshell_open/artifactregistry.go
subfuzion: Good catch; the operation will fail if a repository with the same name already exists, but I agree with the bigger point @grayside is making (I think), which is that it's probably not a good idea to for errors to be ignored at any step (if for no better reason than failing as early as possible is always desirable).

cmd/cloudshell_open/cloudrun.go
glasnt: There's a lot of older code here that I don't think would pass Golang standards today. The above diff is specifically the output of a linting command, which happens to be near one of the many other issues. I won't be addressing all of those in this PR

cmd/cloudshell_open/artifactregistry.go
glasnt: Yup, this makes sense; I was getting confused earlier about the "no new variables" issue, hence the change to explicit declaration. Will address.

cmd/cloudshell_open/artifactregistry.go
```diff
@@ -0,0 +1,69 @@
...
+		return fmt.Errorf("failed to create artifact registry client: %w", err)
+	}
+
+	req := &artifactregistrypb.GetRepositoryRequest{
+		Name: repoFull,
+	}
+	existingRepo, err := client.GetRepository(ctx, req)
+
+	if err != nil {
+		// TODO need to continue if error is "code = NotFound desc = Requested entity was not found."
```
glasnt, line 45: I need to learn how to check for this specific error. 

"code = NotFound desc = Requested entity was not found." is returned if the registry doesn't exist, which is an ok error to get, and should force the creation of the registry.

glasnt: @grayside or @subfuzion any tips? 
This check may replace the `existingRepo == nil` check later (either we allow this "NotFound" error now and continue, or create if this error is encountered)

subfuzion: This might be helpful to you in this situation. Take a look at `errors.Is` (https://pkg.go.dev/errors).

glasnt: I used this, but had to build a grpc error entity to validate against (used https://pkg.go.dev/google.golang.org/grpc/status#Error and https://pkg.go.dev/google.golang.org/grpc/codes)

cmd/cloudshell_open/artifactregistry.go
```diff
@@ -0,0 +1,74 @@
...
+
+		_, err := client.CreateRepository(context.TODO(), req)
+		if err != nil {
+			return fmt.Errorf("failed to create artifact registry: %w", err)
+		}
+	}
+
+	return nil
+
+}
```
NimJay, line 74: Praise: Thanks for writing modular code (e.g., separate file and function for AR logic)!

Dockerfile
```diff
@@ -8,4 +8,4 @@ RUN CGO_ENABLED=0 GOOS=linux go build -a -ldflags '-extldflags "-static"' \
         -o /bin/a.out ./cmd/cloudshell_open
 
 FROM gcr.io/cloudshell-images/cloudshell:latest
-COPY --from=build /bin/a.out /bin/cloudshell_open
\ No newline at end of file
+COPY --from=build /bin/a.out /bin/cloudshell_open
```
NimJay, line 11: **Question** (mostly for my personal understanding, no action required): Does this image need to be manually built and pushed after merging this pull-request?

NimJay (approved):
Looks good to me! Left a few questions. 
Feel free to merge.

**Suggestion:** I also recommend adding a sentence or two about what this pull-request is doing (in the description).

Dockerfile
glasnt: Not manual; there's a Cloud Build Trigger set to build on merge to main for this repo.

cmd/cloudshell_open/main.go
```diff
@@ -291,14 +292,24 @@ func run(opts runOpts) error {
 		}
 	}
 
+	end = logProgress(
+		fmt.Sprintf("Setting up %s in region %s (if it doesn't already exist)", highlight(artifactRegistry), highlight(region)),
+		fmt.Sprintf("Set up %s in region %s (if it doesn't already exist)", highlight(artifactRegistry), highlight(region)),
+		"Failed to setup artifact registry.")
+	err = createArtifactRegistry(project, region, artifactRegistry)
```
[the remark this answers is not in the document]
glasnt: Good point; I need to confirm if artifactregistry is a default API, but always good to make sure it's enabled (even in it's current state, the code doesn't check first then optionally enable, it just always enables). 

Don't think we also need to enable containerregistry any more (and it may not be available on newly created projects)

cmd/cloudshell_open/cloudrun.go
subfuzion: Sorry, conversation isn't resolved yet. I think you thought I was commenting on the lint recommendations. I was commenting on and suggesting that `tryServiceName` actually return an error rather than accept and return an (erroneous) empty string. I realize you didn't touch this code, my suggestion is just based on us leaving code in a better state than we found it.

subfuzion (approved):
LGTM

cmd/cloudshell_open/cloudrun.go
subfuzion: Hi @glasnt, the rest of the PR LGTM and I don't want to be too nitpicky. I asked @telpirion to take a look at this and resolve the conversation accordingly.

PR #289, opened 2024-10-18
chore: batch renovatebot

glasnt:
To prevent trickling renovate PRs (#278, #279, #281, #282, #283, #284, #286, #287, #288), use renovate config (borrowed from https://github.com/GoogleCloudPlatform/cloud-run-microservice-template-go to batch updates weekly.)

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,6 +1,25 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
-    "config:recommended"
-  ]
+    "config:base",
+    "schedule:weekly",
+    ":semanticCommits"
+  ],
+  "postUpdateOptions": [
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -1,6 +1,25 @@
...
+  ],
+  "commitMessagePrefix": "chore(all): ",
+  "commitMessageAction": "update",
+  "groupName": "all",
+  "packageRules": [
+    {
+      "updateTypes": [
+        "major"
+      ],
+      "enabled": false
```
grayside, line 19: If I'm reading this correctly, this is grouping all updates and disabling major updates. It may be grouping all updates except major updates which are left individual, which seems like a better path. Either way seems like a good iteration.

grayside (approved):

PR #342, opened 2026-04-20
fix: explicitly tag and push the latest image

shruti-mantri:


cloudbuild-redirector.yaml
```diff
@@ -1,20 +1,18 @@
...
   - cloud-run-button-deploy
-  - --image=gcr.io/$PROJECT_ID/button-redirector:public-image-${COMMIT_SHA}
+  - --image=gcr.io/$PROJECT_ID/button-redirector:latest
   - --region=us-central1
   - --platform=managed
   - --quiet
   env:
   - CLOUDSDK_CORE_DISABLE_PROMPTS=1
-
-images:
```
HarshaliRaka, line 18: I don't think we should remove the images field, it helps with build provenance. Let's just add :latest as the image tag here

shruti-mantri: This wasn't present before. I had recently added it thinking it was missing.
On deep dive, realized that this was not present because we were explicitly pushing the images in the second step. Because of that, this would be a no-op.

shruti-mantri: Also, looks like the images are explicitly pushed because we need to immediately use them for deploying.

ShellyAgg (approved):

HarshaliRaka (approved):

